PR #24, opened 2018-06-06
fixing logic in getPhase

alex-cory:
I tested this and it works, but need to verify with @dwalintukan if this is the correct logic.


dwalintukan:
your fixes are approved and got merged in with Henry's PR: #25 .
closing this one.

PR #29, opened 2018-06-09
Use shrinkwrap to update WS dependency

dwalintukan:


dwalintukan:
merging. just forcing an update of a WS websocket lib. no code changes.

PR #32, opened 2018-06-12
fix multiple apparent styling issue

amazingandyyy:
fix multiple apparent styling issue and bugs, e.g. undefined [variable](https://github.com/bodhiproject/bodhi-server/pull/32/files#diff-16cf8d8a009f521760bd740d7b6689a1R595) in error messages when I went through the codebase.

please feel free to close or consider to merge.


src/server.js
```diff
@@ -39,7 +37,11 @@ function getQtumProcess() {
 // Checks to see if the qtumd port is still in use
 function checkQtumPort() {
   const port = isMainnet() ? Config.RPC_PORT_MAINNET : Config.RPC_PORT_TESTNET;
-  portscanner.checkPortStatus(port, Config.HOSTNAME, (error, status) => {
+  portscanner.checkPortStatus(port, Config.HOSTNAME, (err, status) => {
+    if (err) {
+      getLogger().error(`Error: qtumd: ${err.message}`);
+      throw err;
```
dwalintukan: remove the throw. This is specifically for when the user uses the `Launch Qtum Wallet` option in the dapp. We check the port continuously to see when we can start the Qtum Wallet since qtumd and qtum-qt can't run simultaneously. throwing here would break the check. let's just fail silently.

src/sync/index.js
```diff
@@ -695,7 +694,7 @@ async function getAddressBalances() {
       const getBotBalancePromises = [];
 
       _.map(addressBatches[loop.iteration()], async (address) => {
-        const getBotBalancePromise = new Promise(async (resolve) => {
+        const getBotBalancePromise = new Promise(async (_resolve) => {
```
dwalintukan: lets name this `getBotBalanceResolve` to be more explicit?

src/server.js
amazingandyyy: I see, make sense.

src/sync/index.js
amazingandyyy: updated

PR #42, opened 2018-06-19
Light Client feature branch

dwalintukan:
**Won't merge this until we are ready for light client, but this will serve as a feature branch for the light client change-over.**

Parse all Events from blockchain in new sync.js (this is a refactored form of the old sync logic and does not handle updating txs)
Added models for each new Event
Added new DBs for each new Event
Separated resolvers.js -> queries.js + mutations.js
Add new queries for new DBs


amazingandyyy:
need to spend time understanding the codebase here, will leave it to @alex-cory or @henryzbw to do a more valid approval.

dwalintukan:
not critical to review at this exact moment, just whenever you guys have time. this wont be ready until `Qrypto` is ready and the server with oauth is done.

PR #43, opened 2018-06-20
Rename utils to index and fix all imports

dwalintukan:


amazingandyyy:
LGTM work as expected.

PR #51, opened 2018-06-20
adding nodemon so we dont have to continuously restart the server on …

alex-cory:
…the b/e when making changes on development


amazingandyyy:
SGTM

PR #53, opened 2018-06-21
52 node mod

dwalintukan:
- Convert server to become node module to be used as lib
- Update qtum client 0.15.2
- Other cleanup


src/index.js
```diff
@@ -1,13 +1,54 @@
 const _ = require('lodash');
 
-const { startServer } = require('./server');
-const { blockchainEnv } = require('./constants');
-const { getDevQtumExecPath } = require('./utils');
+const BodhiServer = require('./server');
```
alex-cory: can we name these as camelCase? Typically we reserve PascalCase for classes

dwalintukan: @alex-cory Isn't this considered a bare object?
https://github.com/airbnb/javascript#naming--PascalCase-singleton

PR #58, opened 2018-07-04
make getting connected nodes work

amazingandyyy:


src/api/network.js
```diff
@@ -2,7 +2,8 @@ const { getInstance } = require('../qclient');
 
 const Network = {
   async getPeerNodeCount() {
-    return getInstance().getPeerInfo().length || 0;
```
dwalintukan: `return await (getInstance().getPeerInfo()).length || 0;`

you need to await `getPeerInfo` since that is returning a `Promise` before you try to access the data (length).

src/api/network.js
```diff
@@ -2,7 +2,8 @@ const { getInstance } = require('../qclient');
...
+    const nodeCount = await getInstance().getPeerInfo();
+    return await nodeCount.length || 0;
```
dwalintukan: see comment above. you don't need to await `nodeCount.length`. that is not an async function.

src/schema/resolvers.js
```diff
@@ -278,6 +278,7 @@ module.exports = {
         addressBalances = await getAddressBalances();
       }
       const peerNodeCount = await network.getPeerNodeCount();
+      console.log(peerNodeCount);
```
dwalintukan: remove

src/api/network.js
amazingandyyy: a mistake lol thanks

PR #60, opened 2018-07-05
fix promising value

amazingandyyy:


amazingandyyy:
should have done this way

PR #61, opened 2018-07-06
Enable wallet unlock + add peerNodeCount param 

dwalintukan:
allows to pass in flag `--passphrase=mypassword` to unlock the wallet at the server level. this is for dev purposes.

### DIFF
--- a/src/config/index.js
+++ b/src/config/index.js
@@ -21,7 +21,7 @@ const Config = {
   DEFAULT_GAS_PRICE: 0.0000004,
   CREATE_CORACLE_GAS_LIMIT: 3500000,
   CREATE_DORACLE_GAS_LIMIT: 1500000,
-  UNLOCK_SECONDS: 86400,
+  UNLOCK_SECONDS: 604800,
 };
 const rpcPassword = getRandomPassword(); // Generate random password for every session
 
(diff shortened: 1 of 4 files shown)


src/sync/index.js
```diff
@@ -162,6 +163,7 @@ async function sync(db) {
             currentBlockCount,
             currentBlockTime,
             await calculateSyncPercent(currentBlockCount, currentBlockTime),
+            await network.getPeerNodeCount(),
```
amazingandyyy: I see.

src/config/index.js
```diff
@@ -21,7 +21,7 @@ const Config = {
   DEFAULT_GAS_PRICE: 0.0000004,
   CREATE_CORACLE_GAS_LIMIT: 3500000,
   CREATE_DORACLE_GAS_LIMIT: 1500000,
-  UNLOCK_SECONDS: 86400,
+  UNLOCK_SECONDS: 604800,
```
amazingandyyy: how we define this?

dwalintukan: It's self defined. I changed it from 1 day to 1 week. This is just to be sure that when the user starts the dapp it wont run out of time.

PR #63, opened 2018-07-09
fix typo

amazingandyyy:
just fix typo when I go through the codebase


amazingandyyy:
it means to be COrocale

Issue #77, opened 2018-08-19
Add `hashId` to Oracle model


dwalintukan:
Add a `hashId` field which is generated from MD5 string, `createdTime` + `name`. We need this to be able to identify and fetch an unconfirmed Oracle as the `approve -> createEvent` txs take place. The `txid` changes in the the unconfirmed Oracle (approve -> create).

1. Create MD5 `hashId` logic
2. Add field to Oracle DB

PR #116, opened 2018-09-11
hashId server change

henryzbw:
HashId change on server side

### DIFF
--- a/src/api/address-manager.js
+++ b/src/api/address-manager.js
@@ -11,8 +11,7 @@ function getContract() {
 
 const AddressManager = {
   async eventEscrowAmount(args) {
-    const {
-      senderAddress, // address
+    const { senderAddress, // address
     } = args;
 
     if (_.isUndefined(senderAddress)) {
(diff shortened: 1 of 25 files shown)


henryzbw:
#77

henryzbw:
@dwalintukan I don't know how to add the hashId in translate to make it work. since i need `this.hashId = `, but i don't know where can i get it to here. i tried to add `this.hashId = hashId`, but the hashId is undefined here. so it caused error.

dwalintukan:
@henryzbw I don't mean to actually add the hashId in there, but you notice how like `name` and `options` for Oracle are null? I meant like a placeholder.

src/api/address-manager.js
```diff
@@ -11,8 +11,7 @@ function getContract() {
 
 const AddressManager = {
   async eventEscrowAmount(args) {
-    const {
-      senderAddress, // address
+    const { senderAddress, // address
```
dwalintukan: hm weird linter change. will have to change the eslint rules later.

PR #123, opened 2018-09-24
Changes for Qrypto Integration

dwalintukan:


src/api/transaction.js
```diff
@@ -31,135 +36,41 @@ const Transaction = {
...
-      throw new TypeError('topicAddress needs to be defined');
+    if ((type === APPROVESETRESULT || type === APPROVEVOTE) && !topicAddress) {
+      throw TypeError('topicAddress needs to be defined');
     }
-    if (type === 'APPROVEVOTE' && !oracleAddress) {
-      throw new TypeError('oracleAddress needs to be defined');
-    }
-
-    // Skip approve if enough allowance
-    let txType = type;
```
henryzbw: why this deleted? since we don't need to do the `approve` and `create/setresult/vote` in two calls?

src/graphql/schema.js
```diff
@@ -47,92 +47,89 @@ enum _TransactionStatus {
...
 }
 
 type Oracle {
   txid: String!
-  version: Int!
   blockNum: Int
-  status: _OracleStatusType!
   address: String
   topicAddress: String
-  resultSetterAddress: String
```
henryzbw: what's the difference between `resultSetterAddress` and `resultSetterQAddress`, and what's the new `resultSetterAddress`?

src/api/transaction.js
dwalintukan: Ya, we only show the Approve Tx for the tx cost now. There is a note for the user prompt that tells them they will have to confirm a follow-up tx.

src/graphql/schema.js
dwalintukan: I removed all `...QAddress` fields since we are storing both as the Qtum address format. eg. `resultSetterAddress` and `resultSetterQAddress` was the same Qtum address.

Issue #11, opened 2018-06-01
[Light Client] Add language field for events


dwalintukan:
Need a `language` field for Topics/Oracles so they can be filtered on the frontend.

henryzbw:
@dwalintukan for this issue, do we need to add the field to smart contract? since i'm still confused when we create an event if the language information is not pushed on chain, how can other persons get the language of this event.

dwalintukan:
@henryzbw Don't need to alter the smart contract. When we do a mutation it will include a language field. The UI does two calls, one for Qrypto and one to the server (mutation).

PR #128, opened 2018-10-02
Language

henryzbw:
#11


src/graphql/schema.js
```diff
@@ -61,6 +61,7 @@ type Topic {
   oracles: [Oracle]
   transactions: [Transaction]
   version: Int!
+  language: String
```
dwalintukan: make this required since we will add a migration script

src/graphql/schema.js
```diff
@@ -84,6 +85,7 @@ type Oracle {
   consensusThreshold: String
   transactions: [Transaction]
   version: Int!
+  language: String
```
dwalintukan: make required

dwalintukan:
FYI just thought of an edge case. This migration will only work with if the DB data is the latest. 

Consider this example:
- DB currently has only parsed up to block 100
- Actual live block is now at 200
- We made the language change live on the UI @ block 150. So only events created at blocks 150+ have the mandatory language field.
- When we start the migration script at the current DB @ block 100, blocks 100-150 will be parsed after migration script has already run so they will not have the db.update applied to them.

I am thinking of this to solve this:
In addition to the migration script, lets add auto insert the language when inserting a Topic/Oracle during the synchronizer **only if the language field does not exist**. This applies to `syncTopicCreated`, `syncCentralizedOracleCreated`, `syncDecentralizedOracleCreated`.
```
// instead of getting the count, lets do a db.find. if result is not null => check if language field exists => if does not exist add en-US language => update or insert topic
            if (await DBHelper.getCount(db.Topics, { txid }) > 0) {
              DBHelper.updateTopicByQuery(db.Topics, { txid }, topic);
              
            } else {
              DBHelper.insertTopic(db.Topics, topic);
            }
```

Let me know if you have any questions.

src/db/index.js
```diff
@@ -19,16 +19,31 @@ const db = {
  * Be sure to wrap each migration script in a try/catch and throw on Error.
  * We don't want the server to run if the migration failed.
  */
-async function applyMigrations() {
+async function applyMigrations(db) {
   // Run migration scripts here
+  const migrationPath = require('path').join(__dirname, 'migrations');
+  const migrations = [];
+  try {
+    await fs.readdirSync(migrationPath).forEach(async (file) => {
```
dwalintukan: shouldn't have to await `readdirSync`. it is synchronous.

src/db/index.js
```diff
@@ -19,16 +19,31 @@ const db = {
...
+      if(file.endsWith('.js')){
```
dwalintukan: add spaces after if and before {

src/db/index.js
```diff
@@ -19,16 +19,31 @@ const db = {
...
+        const migration = await require(`./migrations/${file}`);
```
dwalintukan: `require()` is synchronous too. no need to await.

src/db/index.js
```diff
@@ -19,16 +19,31 @@ const db = {
  * Be sure to wrap each migration script in a try/catch and throw on Error.
  * We don't want the server to run if the migration failed.
  */
-async function applyMigrations() {
+async function applyMigrations(db) {
   // Run migration scripts here
+  const migrationPath = require('path').join(__dirname, 'migrations');
```
dwalintukan: move this inside the try catch as well, it can crash.

src/db/index.js
dwalintukan: we need to ensure after reading the files in /migrations that they are in sorted order as well. we want to make sure 1, 2, 3, etc ordering to ensure migrations run in order.

src/db/index.js
```diff
@@ -19,16 +19,31 @@ const db = {
...
+        const migration = await require(`./migrations/${file}`);
+        migrations.push(migration);
+      }
+    });
+  } catch (err) {
+    throw Error(`Migration scripts load Error ${err.message}`);
+  }
+
+  // execute sequentially
+  migrations.reduce((promise, currentTask) => promise.then(() => currentTask(db)), Promise.resolve()).catch((err) => {
```
dwalintukan: convert to async await syntax

dwalintukan: should be using `map` or `forEach`. 

reduce is meant to run functionality and add to an accumulator: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

src/db/index.js
```diff
@@ -19,16 +19,31 @@ const db = {
...
+    console.error(err);
```
dwalintukan: `getLogger().error(err)`. we output this to Winston logger so we can remotely log this as well.

src/db/migrations/migration1.js
```diff
@@ -0,0 +1,31 @@
...
+
+async function addLanguageField(db) {
+  try {
+    await db.update({ language: { $exists: false } }, { $set: { language: 'en-US' } }, { multi: true });
+  } catch (err) {
+    throw Error(`DB update Error: ${err.message}`);
+  }
+}
+
+
```
dwalintukan: 👾

src/db/migrations/migration1.js
```diff
@@ -0,0 +1,31 @@
+const _ = require('lodash');
+const fs = require('fs-extra');
+const { getLogger } = require('../../utils/logger');
+const Utils = require('../../utils');
+
+async function addLanguageField(db) {
+  try {
+    await db.update({ language: { $exists: false } }, { $set: { language: 'en-US' } }, { multi: true });
+  } catch (err) {
+    throw Error(`DB update Error: ${err.message}`);
```
dwalintukan: also add ```getLogger().error(`DB update Error: ${err.message}`);```
we want to be able to monitor this on Papertrail remote logging.

src/db/migrations/migration1.js
```diff
@@ -0,0 +1,31 @@
...
+  const migrationTrackPath = `${__dirname}/../migrations.dat`;
+  try {
+    let lastMigrate = Number(await fs.readFileSync(migrationTrackPath).toString().split('=')[1].trim());
+    if (lastMigrate < 1) {
+      await addLanguageField(db.Oracles);
+      await addLanguageField(db.Topics);
+      lastMigrate++;
+      getLogger().info(`Success migration${lastMigrate}`);
+    }
+    await fs.outputFileSync(migrationTrackPath, `LAST_MIGRATION=${lastMigrate}\n`);
```
dwalintukan: this logic of writing to `migrations.dat` should be outside of any migration file. move this to `applyMigrations()` since it is generic logic and not specific to the a single migration. otherwise you would have to keep duplicating this code in every migration.

src/db/migrations/migration1.js
```diff
@@ -0,0 +1,31 @@
...
+  } catch (err) {
+    throw Error(`DB update Error: ${err.message}`);
+  }
+}
+
+
+async function migration1(db) {
+  const migrationTrackPath = `${__dirname}/../migrations.dat`;
+  try {
+    let lastMigrate = Number(await fs.readFileSync(migrationTrackPath).toString().split('=')[1].trim());
```
dwalintukan: lets keep track of the `lastMigrate` number in `applyMigrations()` and pass it in as a param to the migration: `migration(db, lastMigrationNum)`

src/db/migrations/migration1.js
```diff
@@ -0,0 +1,31 @@
...
+    if (lastMigrate < 1) {
+      await addLanguageField(db.Oracles);
+      await addLanguageField(db.Topics);
+      lastMigrate++;
+      getLogger().info(`Success migration${lastMigrate}`);
+    }
+    await fs.outputFileSync(migrationTrackPath, `LAST_MIGRATION=${lastMigrate}\n`);
+  } catch (err) {
+    throw Error(`DB track file Error: ${err.message}`);
```
dwalintukan: Not an accurate error message. This catch catches possible crashes in reading the file, addLanguageField, and writing to a file. Try an encompassing generic message that would make sense to the dev seeing the logs: e.g. `Migration 1 error: ${err.message}`

src/sync/index.js
```diff
@@ -59,9 +59,12 @@ const syncTopicCreated = async (currentBlockNum) => {
         topicEventPromises.push(new Promise(async (resolve) => {
           try {
             const topic = new Topic(blockNum, txid, entry).translate();
-
             // Update existing mutated Topic or insert new
             if (await DBHelper.getCount(db.Topics, { txid }) > 0) {
+              const foundTopic = await DBHelper.findOne(db.Topics, { txid }, ['language']);
+              if(foundTopic.language){
```
dwalintukan: add spaces after if and )

src/db/migrations/migration1.js
```diff
@@ -0,0 +1,30 @@
...
+    await db.update({ language: { $exists: false } }, { $set: { language: 'en-US' } }, { multi: true });
+  } catch (err) {
+    getLogger().error(`DB update Error: ${err.message}`);
+    throw Error(`DB update Error: ${err.message}`);
+  }
+}
+
+async function migration1(db, lastMigrate) {
+  try {
+    if (lastMigrate < 1) {
```
dwalintukan: lets check if `lastMigrate === 0` just for safety.

PR #138, opened 2018-10-09
fix #122

### LINKED ISSUE #122: Order Oracles by blockNum DESC in Topics

henryzbw:


src/graphql/resolvers.js
```diff
@@ -26,7 +26,7 @@ module.exports = {
   Mutation: Mutations,
 
   Topic: {
-    oracles: ({ address }, data, { db: { Oracles } }) => Oracles.find({ topicAddress: address }),
+    oracles: ({ address }, data, { db: { Oracles } }) => Oracles.cfind({ topicAddress: address }).sort({blockNum: -1}).exec(),
```
dwalintukan: why do you need to change it to the async find call? does `Oracles.find({ topicAddress: address }).sort({blockNum: -1}).exec()` not work?

dwalintukan: also, add spacing for the obj with blockNum

henryzbw: we are using `nedb-promise` which is a thin wrapper around nedb, so the cursors are mapped to all the cfunctions, so we need to use `cfind` to get the cursor and then set the sorting parameter.

dwalintukan: But this was working before using `find`. The wrapper allows for using both the sync and async functions.

henryzbw: using `find` will directly return a promise, which is not allowed to set the order.  sort is implemented inside Cursor.

generally db.find(query)  = db.cfind(query).exec(), which the cursor returned by cfind can be used to set `sort, skip and limit`

dwalintukan: gotcha. hm good to know, so `find` which is normally sync turns into async when using the promisified wrapper.

PR #141, opened 2018-10-16
add mostvoted

henryzbw:
Server change for mostvoted

### DIFF
--- a/src/graphql/queries.js
+++ b/src/graphql/queries.js
@@ -418,24 +418,22 @@ module.exports = {
     return cursor.exec();
   },
 
-  mostVotes: async (root, { filter,orderBy, limit, skip }, { db: { Votes } }) => {
+  mostVotes: async (root, { filter, orderBy, limit, skip }, { db: { Votes } }) => {
     const query = filter ? { $or: buildVoteFilters(filter) } : {};
-    let cursor = Votes.cfind(query);
+    const cursor = Votes.cfind(query);
     const result = await cursor.exec();
 
-    const accumulated = result.reduce((acc, cur) =>{
-      if(acc.hasOwnProperty(cur.voterAddress)){
+    const accumulated = result.reduce((acc, cur) => {
+      if (acc.hasOwnProperty(cur.voterAddress)) {
         acc[cur.voterAddress] += Number(cur.amount);
-      }else{
+      } else {
         acc[cur.voterAddress] = Number(cur.amount);
       }
       return acc;
-    },{});
+    }, {});
 
-    let votes = Object.keys(accumulated).map((key) => {
-      return {voterAddress: key, amount: String(accumulated[key])};
-    })
-    votes.sort((a,b) => b.amount - a.amount);
+    let votes = Object.keys(accumulated).map(key => ({ voterAddress: key, amount: String(accumulated[key]) }));
+    votes.sort((a, b) => b.amount - a.amount);
 
     const totalCount = votes.length;
     let hasNextPage;
--- a/src/graphql/queries.js
+++ b/src/graphql/queries.js
@@ -1,6 +1,5 @@
 const _ = require('lodash');
 const BigNumber = require('bignumber.js');
-const axios = require('axios');
 const { calculateSyncPercent } = require('./subscriptions');
 const { getInstance } = require('../qclient');
 const { SATOSHI_CONVERSION, STATUS, TOKEN } = require('../constants');
@@ -10,6 +9,7 @@ const Blockchain = require('../api/blockchain');
 const Wallet = require('../api/wallet');
 const BodhiToken = require('../api/bodhi-token');
 const Network = require('../api/network');
+const TopicEvent = require('../api/topic-event');
 
 const DEFAULT_LIMIT_NUM = 50;
 const DEFAULT_SKIP_NUM = 0;
@@ -335,6 +335,21 @@ const getAddressBalances = async () => {
   return addressObjs;
 };
 
+const getWinnings = async (vote) => {
+  const data = await TopicEvent.calculateWinnings({
+    contractAddress: vote.topicAddress,
+    senderAddress: vote.voterAddress,
+  });
+  const win = {};
+  win.topicAddress = vote.topicAddress;
+  win.voterAddress = vote.voterAddress;
+  const amount = {};
+  amount.bot = data[0];
+  amount.qtum = data[1];
+  win.amount = amount;
+  return win;
+};
+
 module.exports = {
   allTopics: async (root, { filter, orderBy, limit, skip }, { db: { Topics } }) => {
     const query = filter ? { $or: buildTopicFilters(filter) } : {};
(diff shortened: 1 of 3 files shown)


src/graphql/queries.js
```diff
@@ -415,6 +418,47 @@ module.exports = {
     return cursor.exec();
   },
 
+  mostVotes: async (root, { filter, orderBy, limit, skip }, { db: { Votes } }) => {
```
dwalintukan: we should validate that a `topicAddress` is passed in.

src/graphql/schema.js
```diff
@@ -188,6 +188,18 @@ type PaginatedTopics {
   pageInfo: PageInfo
 }
 
+
+type AccumulatedVotes {
+  voterAddress: String!
+  amount: String!
+}
+
+type PaginatedVotes {
```
dwalintukan: change to `PaginatedAccumulatedVotes`. we should be consistent and have `PaginatedVotes` on the Vote table

src/graphql/schema.js
```diff
@@ -188,6 +188,18 @@ type PaginatedTopics {
   pageInfo: PageInfo
 }
 
+
+type AccumulatedVotes {
```
dwalintukan: add `topicAddress`

src/graphql/queries.js
```diff
@@ -415,6 +418,48 @@ module.exports = {
...
+    const voterFilters = buildVoteFilters(filter);
+    voterFilters.forEach((element) => {
+      if (element.topicAddress == null) {
+        throw new Error('topicAddress is not passed in');
```
dwalintukan: nit: `topicAddress is required`

src/graphql/queries.js
```diff
@@ -415,6 +418,48 @@ module.exports = {
...
+      }
+    });
+    const query = filter ? { $or: voterFilters } : {};
+    const cursor = Votes.cfind(query);
+    const result = await cursor.exec();
```
dwalintukan: nit: combine `cursor` and `result`

src/graphql/queries.js
```diff
@@ -415,6 +418,48 @@ module.exports = {
     return cursor.exec();
   },
 
+  mostVotes: async (root, { filter, orderBy, limit, skip }, { db: { Votes } }) => {
+    const voterFilters = buildVoteFilters(filter);
```
dwalintukan: think we should only allow max 1 filter in the `filter`. so we don't data from two events. add a check to make sure the length === 1.

2 checks:
length === 1
topicAddress is there

src/graphql/queries.js
dwalintukan: also, fyi: `throw Error()` for shorthand

src/graphql/queries.js
```diff
@@ -1,9 +1,9 @@
 const _ = require('lodash');
 const BigNumber = require('bignumber.js');
-
+const axios = require('axios');
```
dwalintukan: remove

src/graphql/schema.js
```diff
@@ -188,6 +188,26 @@ type PaginatedTopics {
...
+  topicAddress: String!
+  voterAddress: String!
+  amount: String!
```
dwalintukan: we should return the `token` as well

src/graphql/schema.js
```diff
@@ -188,6 +188,26 @@ type PaginatedTopics {
...
+type PaginatedAccumulatedVotes {
+  totalCount: Int!
+  votes: [AccumulatedVotes]!
+  pageInfo: PageInfo
+}
+
+type Winners {
+  topicAddress: String!
+  voterAddress: String!
+  amount: String!
```
dwalintukan: we need to show the amounts for both tokens:
```
qtumAmount: String!
botAmount: String!
```

src/graphql/schema.js
```diff
@@ -188,6 +188,26 @@ type PaginatedTopics {
...
+  token: _TokenType!
```
dwalintukan: can be removed once adding both amounts

src/graphql/schema.js
dwalintukan: Singular form: `AccumulatedVote`

src/graphql/schema.js
```diff
@@ -188,6 +188,26 @@ type PaginatedTopics {
...
+  amount: String!
+}
+
+type PaginatedAccumulatedVotes {
+  totalCount: Int!
+  votes: [AccumulatedVotes]!
+  pageInfo: PageInfo
+}
+
+type Winners {
```
dwalintukan: Singluar: `type Winner`.

src/graphql/queries.js
```diff
@@ -415,6 +418,82 @@ module.exports = {
...
+        filtered.push(vote);
+      }
+    });
+    let winnings = [];
+
+    for (const item of filtered) {
+      winnings.push( await (a(item)));
+    }
+    winnings = _.orderBy(winnings,['amount'], ['desc'] )
+    async function a (item)  {
```
dwalintukan: couple of comments:
1. either remove this function entirely and just call `TopicEvent.calculateWinnings()` as the API does in the for loop. or move this function outside of the `module.exports` so it can be reused.
2. it's good practice to name it something that you can tell without having to read the code like `getWinnings()`

src/graphql/queries.js
```diff
@@ -415,6 +418,82 @@ module.exports = {
...
+        voterAddress: vote.voterAddress,
+        topicAddress: vote.topicAddress,
+      })) {
+        filtered.push(vote);
+      }
+    });
+    let winnings = [];
+
+    for (const item of filtered) {
+      winnings.push( await (a(item)));
```
dwalintukan: fix spacing

src/graphql/queries.js
```diff
@@ -415,6 +418,82 @@ module.exports = {
...
+    winnings = _.orderBy(winnings,['amount'], ['desc'] )
+    async function a (item)  {
+      const { data } = await axios.post('http://localhost:6767/winnings', { // eslint-disable-line
+      contractAddress: item.topicAddress,
+      senderAddress: item.voterAddress,
+      });
+      const win = {};
+      win.topicAddress = item.topicAddress;
+      win.voterAddress = item.voterAddress;
+      win.amount = filter.OR[0].token === TOKEN.BOT ? data[0] : data[1];
```
dwalintukan: return both `qtumAmount`  and `botAmount`

src/graphql/queries.js
```diff
@@ -415,6 +418,82 @@ module.exports = {
...
+      win.token = filter.OR[0].token;
```
dwalintukan: remove

src/graphql/queries.js
```diff
@@ -332,6 +335,21 @@ const getAddressBalances = async () => {
...
+    senderAddress: vote.voterAddress,
+  });
+  const win = {};
+  win.topicAddress = vote.topicAddress;
+  win.voterAddress = vote.voterAddress;
+  const amount = {};
+  amount.bot = data[0];
+  amount.qtum = data[1];
+  win.amount = amount;
+  return win;
```
dwalintukan: nit: why not just 
```
return {
  topicAddress: vote.topicAddress,
  voterAddress: vote.voterAddress,
  amount: {
    bot: data[0],
    qtum: data[1],
  }
};
```

src/graphql/schema.js
```diff
@@ -191,6 +191,29 @@ type PaginatedTopics {
...
+  votes: [AccumulatedVote]!
+  pageInfo: PageInfo
+}
+
+type Winner {
+  topicAddress: String!
+  voterAddress: String!
+  amount: TokenAmount!
+}
+type TokenAmount {
```
dwalintukan: nice 👍

Issue #144, opened 2018-10-18
votes.db migration


BernardHan:
add a `type` field in `votes.db`: [VOTE, RESULT_SET, BET]

@dwalintukan 
By adding this field, I think we can just pull resultsetting history from `votes.db` not `resultsets.db`, since then we can tell if it's a result set and it has `amount`. There's no need to migrate `resultsets.db` anymore.

Implementation:
1. go through all `votes`
2. if the entry uses `QTUM`, then it's a `BET`
3. if it uses `BOT`, then query the Oracle
4. if Oracle == Centralized, then it's a `RESULT_SET`, otherwise it's a `VOTE`

BernardHan:
Or maybe I don't even need to do migration?

Like how you added `block`:

``` js
  Vote: {
    block: async ({ blockNum }, data, { db: { Blocks } }) => (await Blocks.find({ blockNum }))[0],
    type:  async ({ token, oracleAddress }, { db: { Oracles } }) => {
        if (token === 'QTUM') return 'BET';
        const oracle = await Oracles.find({ oracleAddress });
        if (oracle.token === 'QTUM') return 'RESULT_SET';
        return 'VOTE';
    },
  },
```

dwalintukan:
@BernardHan this could work by adding a resolver. But at the cost of increased query time. I think the migration makes more sense long term to have it as part of the table.

BernardHan:
@dwalintukan I'll finish the migration tomorrow.

dwalintukan:
Sounds good.

PR #145, opened 2018-10-19
added field type into vote.db

BernardHan:
#144


src/db/migrations/migration2.js
```diff
@@ -0,0 +1,33 @@
...
+const { map } = require('lodash');
+
+async function addTypeField(db) {
+  try {
+    // await db.update({ type: { $exists: false }, token: 'QTUM' }, { $set: { type: 'BET' } }, { multi: true });
+    // await db.update({ type: { $exists: false }, token: 'QTUM' }, { $set: { type: 'BET' } }, { multi: true });
+    const votes = await db.Votes.find({ type: { $exists: false } });
+    map(votes, async (vote) => {
+      if (vote.token === 'QTUM') {
+        await db.Votes.update({ txid: vote.txid }, { $set: { type: 'BET' } });
```
dwalintukan: create a var for `type` and call `await db.Votes.update` once at the end of the if checks

src/db/migrations/migration2.js
```diff
@@ -0,0 +1,33 @@
...
+    });
+  } catch (err) {
+    getLogger().error(`DB update Error: ${err.message}`);
+    throw Error(`DB update Error: ${err.message}`);
+  }
+}
+
+async function migration2(db) {
+  try {
+    await addTypeField(db);
```
dwalintukan: inline the func here

src/db/migrations/migration2.js
```diff
@@ -0,0 +1,33 @@
...
+        await db.Votes.update({ txid: vote.txid }, { $set: { type: 'BET' } });
+      } else if (vote.token === 'BOT') {
+        const oracle = await db.Oracles.find({ address: vote.oracleAddress });
+        if (oracle[0].token === 'QTUM') await db.Votes.update({ txid: vote.txid }, { $set: { type: 'RESULT_SET' } });
+        else if (oracle[0].token === 'BOT') await db.Votes.update({ txid: vote.txid }, { $set: { type: 'VOTE' } });
+      }
+    });
+  } catch (err) {
+    getLogger().error(`DB update Error: ${err.message}`);
+    throw Error(`DB update Error: ${err.message}`);
```
dwalintukan: lets be more specific for the error message. something like `Migration 2 error: ${err.message}`

src/db/migrations/migration2.js
```diff
@@ -0,0 +1,33 @@
...
+async function migration2(db) {
+  try {
+    await addTypeField(db);
+  } catch (err) {
+    getLogger().error(err.message);
+    throw err;
+  }
+}
+
+module.exports = migration2;
```
dwalintukan: lets add a test case for the migration:
1. create a test db
2. add 2 oracles, 1 centralized, 1 decentralized
3. add 3 votes, 1 for each type
4. run the migration on this test db.
5. validate votes now have the type field

just copy some data from the dev graphql db

src/sync/index.js
```diff
@@ -199,6 +199,12 @@ const syncOracleResultVoted = async (currentBlockNum) => {
             // Add Topic address to Vote
             const oracle = await DBHelper.findOne(db.Oracles, { address: vote.oracleAddress });
             vote.topicAddress = oracle.topicAddress;
+            if (vote.token === 'QTUM') {
+              vote.type = 'BET';
```
dwalintukan: add the `VOTE_TYPE` constants in `constants.js` and use here

src/db/index.js
```diff
@@ -14,7 +14,7 @@ const db = {
   Blocks: undefined,
   Transactions: undefined,
 };
-const MIGRATION_REGEX = /(migration)(\d+)/g;
```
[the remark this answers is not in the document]
BernardHan: So if we are looking for multiple occurrences of that pattern among a string, `g` would be a must. In our case we just checking if the whole string follows the pattern, `g` is not necessary.

But more importantly, by having `g`, there exists an inner index counter for the purpose of multiple matches. But since we declare it as global, the counter won't have a chance to reset, which affects further searches in a loop.

We could still have `g`, but we need to put the declaration inside the loop. I removed it because it's an overkill.

src/db/index.js
```diff
@@ -74,9 +74,9 @@ async function applyMigrations() {
   /* eslint-disable no-restricted-syntax, no-await-in-loop */
   for (const migration of migrations) {
     try {
-      if (migration.number === lastMigration + 1) {
+      if (migration.number == lastMigration + 1) { // eslint-disable-line
```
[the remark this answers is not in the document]
BernardHan: `migration.number` is type of `string`, while `lastMigration` is type of number.

src/db/migrations/migration2.js
BernardHan: why inline?

src/db/index.js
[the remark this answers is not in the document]
dwalintukan: Gotcha. Ya I remember reading about this before.

src/db/index.js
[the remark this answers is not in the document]
dwalintukan: lets cast `migration.number` to a number then

src/db/migrations/migration2.js
dwalintukan: we only call `addTypeField()` once so what is the purpose of a `migration2()` calling another func? `addTypeField` is also already wrapped in a try/catch.

BernardHan: Gotcha

BernardHan: Gotcha

BernardHan: Gotcha

src/db/migrations/migration2.js
BernardHan: What I did to test was 
1. copy the `votes.db` into a backup, then directly run migration on `votes.db`.
2. I check and compare the original number of entries with the number of `type` occurs, which is the same. 
3. Then I check if the type is correct by searching `oracle address` in `oracles.db`, after checking about 10 entries, I concluded that the result is what the code instructed.

Do you want me to do the test you mention here?

dwalintukan: I was referring to actual unit tests. Although your's test does work, its better to have repeatable unit tests so that you have test coverage for that specific area. 

For example, the migration works right now, but you don't have any unit tests for it. Then later down the line some code changes that breaks the migration. Then anyone running the migration from that point forward will have a broken server. Thus having test coverage on the migration would pick this up when someone broke the migration code.

Can you add the test cases I mentioned above?

BernardHan: Will do it!

BernardHan: I tried to add test.
1. I create mock db as a file in mock folder
2. votes.db is named into votes.copy
3. every time when running the test, it first removes existing `mock/votes`, then copy `mock/votes.copy` into `mock/votes`, then run migration on `mock/votes`
4. I have `initDB` based on the mock db in my test
5. but the migration did not work, need to spend more time on searching for reason, after migration works, I plan to `streamRead` the result file to check the `type`

dwalintukan: @BernardHan When adding the entries into the tables do it the native way:
1. `initDB()` -> create new db files: `votes.test.db` & `oracles.test.db` -> loadDatabase -> etc
2. `DBHelper.insertOracle()` x2 - centralized and decentralized
3. `db.Votes.insert(vote)` x3 - one for each type
4. run migration test
5. tests pass
6. delete `votes.test.db` & `oracles.test.db` with `fs.unlink()`

src/db/migrations/test/migration2.tests.js
```diff
@@ -0,0 +1,146 @@
...
+}
+
+describe('Type Field Migration', () => {
+  describe('1 central & 1 decentral oracle + votes for each type', () => {
+    const db = {
+      Oracles: undefined,
+      Votes: undefined,
+    };
+
+    it('Run migration', async () => {
```
dwalintukan: nit: `it('completes the migration', async () => {`

src/db/migrations/test/migration2.tests.js
```diff
@@ -0,0 +1,146 @@
...
+    topicAddress: '64a2cc9a78f2faff6a7401e393f3a90731704ab6',
+  };
+  await db.Votes.insert(resultset);
+  await db.Votes.insert(vote);
+  await db.Votes.insert(bet);
+}
+
+async function initDB(db) {
+  if (fs.existsSync(path.join(__dirname, './mock/votes.test.db'))) fs.unlinkSync(path.join(__dirname, './mock/votes.test.db'));
+  if (fs.existsSync(path.join(__dirname, './mock/oracles.test.db'))) fs.unlinkSync(path.join(__dirname, './mock/oracles.test.db'));
```
dwalintukan: lets delete the test DBs after the test is done so we dont have the files there.

PR #146, opened 2018-10-24
Votes parsing fix

dwalintukan:


henryzbw:
tested it, bet 3 times still only add 1 to the db, seems not fixed

PR #149, opened 2018-10-26
support leaderboard stats

henryzbw:


src/graphql/queries.js
```diff
@@ -438,17 +438,16 @@ module.exports = {
...
-    }
+
     const query = filter ? { $or: voterFilters } : {};
     const result = await Votes.find(query);
 
     const accumulated = result.reduce((acc, cur) => {
+      const curAmount = new BigNumber(cur.amount);
       if (acc.hasOwnProperty(cur.voterAddress)) {
-        acc[cur.voterAddress] += Number(cur.amount);
+        acc[cur.voterAddress] = new BigNumber(acc[cur.voterAddress]).plus(curAmount).toString(10);
```
dwalintukan: nice BigNumber usage 👍

src/graphql/queries.js
```diff
@@ -496,6 +495,32 @@ module.exports = {
     return winnings;
   },
 
+  leaderboardStats: async (root, { filter, orderBy, limit, skip }, { db: { Votes, Topics } }) => {
+      const result = await Votes.find({});
```
dwalintukan: we dont need to specify a high limit to get all the items?

src/graphql/queries.js
```diff
@@ -496,6 +495,32 @@ module.exports = {
...
+      let totalBOT = new BigNumber(0);
+      result.reduce((acc, cur) => {
+        const curAmount = new BigNumber(cur.amount);
+        if (!acc.hasOwnProperty(cur.voterAddress)) {
+          acc[cur.voterAddress] = '';
+          participantsCount++;
+        }
+        if (cur.token === TOKEN.BOT) {
+          totalBOT = new BigNumber(totalBOT).plus(curAmount).toString(10);
+        }else{
```
dwalintukan: spacing

src/graphql/queries.js
```diff
@@ -496,6 +495,32 @@ module.exports = {
...
 
+  leaderboardStats: async (root, { filter, orderBy, limit, skip }, { db: { Votes, Topics } }) => {
+      const result = await Votes.find({});
+      let participantsCount = 0;
+      let totalQTUM = new BigNumber(0);
+      let totalBOT = new BigNumber(0);
+      result.reduce((acc, cur) => {
+        const curAmount = new BigNumber(cur.amount);
+        if (!acc.hasOwnProperty(cur.voterAddress)) {
+          acc[cur.voterAddress] = '';
```
dwalintukan: lets just init with new BigNumber(0)

src/graphql/queries.js
```diff
@@ -496,6 +495,32 @@ module.exports = {
...
+          participantsCount++;
+        }
+        if (cur.token === TOKEN.BOT) {
+          totalBOT = new BigNumber(totalBOT).plus(curAmount).toString(10);
```
dwalintukan: keep it as a BigNumber type the whole time so you dont have to convert toString on every iteration. then on the return you can `totalQTUM.toString(10)` and `totalBOT.toString(10)`.

src/graphql/queries.js
```diff
@@ -496,6 +495,32 @@ module.exports = {
...
+        }else{
+          totalQTUM = new BigNumber(totalQTUM).plus(curAmount).toString(10);
```
dwalintukan: ditto

src/graphql/queries.js
```diff
@@ -438,17 +438,16 @@ module.exports = {
     if (voterFilters.length !== 1) {
       throw Error('only one event is allowed');
     }
-    if (voterFilters[0].topicAddress == null) {
```
dwalintukan: why remove the check?

src/graphql/schema.js
```diff
@@ -193,7 +193,7 @@ type PaginatedTopics {
 
 
 type AccumulatedVote {
-  topicAddress: String!
+  topicAddress: String
```
dwalintukan: why remove the required `topicAddress`?

src/graphql/schema.js
```diff
@@ -210,6 +210,14 @@ type Winner {
   voterAddress: String!
   amount: TokenAmount!
 }
+
+type LeaderboardStats {
+  eventCount: String!
+  participantsCount: String!
+  totalQTUM: String!
```
dwalintukan: nit: totalQtum

src/graphql/schema.js
```diff
@@ -210,6 +210,14 @@ type Winner {
...
+  totalBOT: String!
```
dwalintukan: nit: totalBot

src/graphql/queries.js
henryzbw: since when we do global leaderboard, we don't need specific topicAddress

src/graphql/schema.js
henryzbw: for the global leaderboard, we don't have topicAddress passed in, so it will be null

src/graphql/queries.js
henryzbw: the 50 default limit is only applied when we use `buildCursorOptions`, if we don't use cursor builder, we won't have a limit set.

Issue #157, opened 2019-05-08
[Naka] Update GraphQL schema, queries, mutations, subscriptions


dwalintukan:
Match the contract

PR #160, opened 2019-05-11
158 sync logic

dwalintukan:
#153 
#156 
#157 
#158


src/db/index.js
```diff
@@ -1,53 +1,96 @@
...
+const { logger } = require('../utils/logger');
 
-const Utils = require('../utils');
-const { getLogger } = require('../utils/logger');
-
+const MIGRATION_REGEX = /(migration)(\d+)/;
 const db = {
-  Topics: undefined,
-  Oracles: undefined,
-  Votes: undefined,
```
henryzbw: We don't have Votes db now?

dwalintukan: Votes -> Bets

Issue #166, opened 2019-05-16
Add Transaction info on mutation based on txid


dwalintukan:
call `getTransaction` on mutation

PR #168, opened 2019-05-16
167 confirm pending

dwalintukan:
#167 #166 
- added `includeRoundBets` param for `events` query to return round balances


src/utils/web3-utils.js
```diff
@@ -0,0 +1,50 @@
+const { isNull } = require('lodash');
+const web3 = require('../web3');
```
henryzbw: const  { web3 } = require('../web3');

src/utils/web3-utils.js
```diff
@@ -0,0 +1,50 @@
...
+const TransactionReceipt = require('../models/tx-receipt');
+
+/**
+ * Gets the pending tx and conforms it to the TransactionReceipt model.
+ * @param {string} txid Transaction ID
+ * @return {TransactionReceipt} Transaction receipt
+ */
+const getTransaction = async (txid) => {
+  const tx = web3().eth.getTransaction(txid);
```
henryzbw: need to await

src/utils/web3-utils.js
```diff
@@ -0,0 +1,50 @@
...
+  return new TransactionReceipt({
+    transactionHash: tx.hash,
+    blockHash: tx.blockHash,
+    blockNumber: tx.blockNumber,
```
henryzbw: tx.blockNumber is null when i get it
TransactionReceipt validation at this part has problem. see the tx returned from `web3().eth.getTransaction(txid);` 
image

src/utils/web3-utils.js
dwalintukan: fixed

src/utils/web3-utils.js
dwalintukan: fixed

src/utils/web3-utils.js
dwalintukan: fixed

PR #173, opened 2019-05-17
Fix multiple filter not working

BernardHan:
`let filters = filter ? [filter] : [];`
When filter is `{}`, filters always add a `{}` into the whole filter, which will query all event


dwalintukan:
@BernardHan Updated all the filters to account for the OR filters in #172

PR #189, opened 2019-05-24
Remove dup

BernardHan:


dwalintukan:
@BernardHan We don't want to actually remove the items from the db. That would not mirror whats emitted from the chain.

The only part this is affecting is the `transactions` query so we just need to manually remove the duplicate from that query only before its returned to the user.

PR #201, opened 2019-05-31
Fix sync error

henryzbw:
UnhandledPromiseRejectionWarning: Error: Error syncMultipleResultsEventCreated: Node error: {"code":-32602,"message":"invalid argument 0: hex string without 0x prefix"}
    at syncMultipleResultsEventCreated (/Users/Bowen/Desktop/bodhi-server/src/sync/multiple-results-event-created.js:62:11)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:160:7)
(node:44320) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:44320) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

@dwalintukan 
When i create event, it will throw above error, the adjustStartBlock is 0.
If I convert the blocknum to hex string, it throws block number out of range , so i just set min to the startBlock, not sure if its right.


dwalintukan:
@henryzbw ya my logic for finding pending was flawed to begin with. i will change the logic in a separate PR.

dwalintukan:
@henryzbw handled in #200

PR #202, opened 2019-05-31
Filter votes from resultsets

BernardHan:


src/graphql/queries/transactions.js
```diff
@@ -77,8 +82,8 @@ module.exports = async (
   // Run ResultSets query
   cursor = ResultSets.cfind(buildResultSetFilters(txFilters));
   cursor = buildCursorOptions(cursor, orderBy, limit);
-  const resultSets = await cursor.exec();
-
+  let resultSets = await cursor.exec();
+  resultSets = filterResultSets(resultSets);
```
dwalintukan: instead of filtering manually, add the filter in `buildResultSetFilters`. try to use the native way (in this case, part of the db query) whenever possible.

PR #207, opened 2019-06-05
Handle events query change

henryzbw:


src/graphql/schema.js
```diff
@@ -291,6 +298,8 @@ type Query {
     skip: Int
     pendingTxsAddress: String
     includeRoundBets: Boolean
+    userAddress: String
```
dwalintukan: lets name this something more descriptive: `roundBetsAddress`

src/graphql/resolvers.js
```diff
@@ -70,29 +70,58 @@ module.exports = {
           eventRound: currentRound,
         });
 
+
         // Sum all bets by index
         const rounds = fill(Array(numOfResults), '0');
+        const userRound = fill(Array(numOfResults), '0');
         each(bets, (bet) => {
+          if (userAddress && userAddress.toLowerCase() === bet.betterAddress) {
```
dwalintukan: i created a helper `toLowerCase` method in `utils/index.js`. please use that one.

src/graphql/resolvers.js
```diff
@@ -70,29 +70,58 @@ module.exports = {
...
+        const userBetRound = fill(Array(numOfResults), '0');
+        if (includeBetRoundBets) {
+          const bets = await DBHelper.findBet({
+            txStatus: TX_STATUS.SUCCESS,
+            eventAddress: address,
+            eventRound: 0,
+          });
+
+          each(bets, (bet) => {
+            if (userAddress && userAddress.toLowerCase() === bet.betterAddress) {
```
dwalintukan: same, use helper toLowerCase

src/graphql/resolvers.js
dwalintukan: and technically wont need to check if `userAddress` is defined. `bet.betterAddress` should always be defined. so if userAddress was undefined, youd be checking `undef === bet.betterAddress` which will fail anyways.

src/graphql/resolvers.js
```diff
@@ -70,29 +70,58 @@ module.exports = {
...
         // Add result set amount if round 1
         if (currentRound === 1) {
           const resultSet = await DBHelper.findOneResultSet({
             txStatus: TX_STATUS.SUCCESS,
             eventAddress: address,
             eventRound: 0,
           });
+
           if (!isNull(resultSet)) {
+            if (userAddress && userAddress.toLowerCase()  === resultSet.centralizedOracleAddress) {
```
dwalintukan: use toLowerCase helper

src/graphql/resolvers.js
```diff
@@ -70,29 +70,58 @@ module.exports = {
...
+            }
             rounds[resultSet.resultIndex] = sumBN(
               rounds[resultSet.resultIndex],
               resultSet.amount,
             ).toString(10);
           }
         }
-
-        return rounds;
+        return {totalRoundBets: rounds, userRoundBets: userRound, totalBetRoundBets: betRound, userBetRoundBets: userBetRound};
```
dwalintukan: nit: place key/value on separate lines

henryzbw:
> Ship after changes

need to go with frontend, otherwise may cause some error.

PR #210, opened 2019-06-06
Transaction Loadmore

BernardHan:
#205 

So since we are changing the history table into something like twitter notification, we need to implement transaction history `loadmore` to make it elegent instead of pull 500 entries at one time.

I'm passing a `skip` parameter for each db when querying `transaction`, and return the `nextSkip` so that the UI know where to query the next time.

We cannot use a general skip, for example:
we want to get 20 txs, so we query `20 events`, `20 bets`, `20 resultsets`, `20 withdraws`, and then we only take 20 items from all of them. If the skip is not specific for each db, some items will be missed on the next query starting from the general skip.

If we don't make a generic transaction table, this is the best solution I can come up with.

### DIFF
--- a/src/graphql/schema.js
+++ b/src/graphql/schema.js
@@ -135,6 +135,7 @@ type Bet implements Transaction {
   amount: String!
   eventRound: Int!
   resultName: String
+  eventName: String
 }
 
 type PaginatedBets {
@@ -158,6 +159,7 @@ type ResultSet implements Transaction {
   nextConsensusThreshold: String
   nextArbitrationEndTime: Int
   resultName: String
+  eventName: String
 }
 
 type PaginatedResultSets {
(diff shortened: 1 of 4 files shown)


src/graphql/schema.js
```diff
@@ -66,6 +66,14 @@ type PageInfo {
   hasNextPage: Boolean!
   pageNumber: Int!
   count: Int!
+  nextSkips: NextSkips
```
dwalintukan: lets call this `nextTransactionSkip`. dont think this will need to be used by another query.

src/graphql/schema.js
```diff
@@ -66,6 +66,14 @@ type PageInfo {
...
+}
+
+type NextSkips {
```
dwalintukan: NextTransactionSkip

src/graphql/schema.js
```diff
@@ -342,6 +357,7 @@ type Query {
     filter: TransactionFilter
     limit: Int
     skip: Int
+    skips: TransactionSkip
```
dwalintukan: `transactionSkip: TransactionSkip`

PR #211, opened 2019-06-06
Change total result bets

henryzbw:
Change for getting:
totalBets,
betterBets,
totalVotes,
betterVotes,

then will use them to do 
IMG_6673

implement the withdraw table

### DIFF
--- a/src/graphql/queries/total-result-bets.js
+++ b/src/graphql/queries/total-result-bets.js
@@ -42,11 +42,11 @@ module.exports = async (
 
   // Get result set
   const resultSet = await DBHelper.findOneResultSet({ eventAddress, eventRound: 0, txStatus: TX_STATUS.SUCCESS });
-  const { centralizedOracleAddress } = resultSet;
+  const { centralizedOracleAddress } = resultSet || {};
 
   // Accumulate all result bets
   let totalVoteRound = await DBHelper.findBet({ eventAddress, eventRound: { $gt: 0 }, txStatus: TX_STATUS.SUCCESS } );
-  totalVoteRound = totalVoteRound.concat(resultSet);
+  if (resultSet) totalVoteRound = totalVoteRound.concat(resultSet);
 
   // Add result set amount
   const totalVotes = accumulateBets(numOfResults, totalVoteRound);
--- a/src/graphql/queries/total-result-bets.js
+++ b/src/graphql/queries/total-result-bets.js
@@ -2,6 +2,7 @@ const { isNull, fill, each, map } = require('lodash');
 const { lowercaseFilters } = require('./utils');
 const DBHelper = require('../../db/db-helper');
 const web3 = require('../../web3');
+const { TX_STATUS } = require('../../constants');
 const { toLowerCase } = require('../../utils/index');
 
 const accumulateBets = (numOfResults, bets) => {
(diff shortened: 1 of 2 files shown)


src/graphql/queries/total-result-bets.js
```diff
@@ -27,19 +27,40 @@ module.exports = async (
...
   const { numOfResults } = event;
 
+  // Accumulate all round 0 bets
+  let totalBetRound = await DBHelper.findBet({ eventAddress, eventRound: 0 });
+  const totalBets = accumulateBets(numOfResults, totalBetRound);
+
+  // Accumulate better round 0 bets
+  let betterBets
+  if (betterAddress) {
+    betterBetRound = await DBHelper.findBet({ eventAddress, eventRound: 0, betterAddress });
```
dwalintukan: whats the diff between this and `roundBets` in the event query? seems the same as that except that its all the votes instead of just the current round.

src/graphql/queries/total-result-bets.js
```diff
@@ -27,19 +27,40 @@ module.exports = async (
...
+  const { resultIndex, amount: resultSetAmount, centralizedOracleAddress } = resultSet;
+
   // Accumulate all result bets
-  let bets = await DBHelper.findBet({ eventAddress });
-  const resultBets = accumulateBets(numOfResults, bets);
+  let totalVoteRound = await DBHelper.findBet({ eventAddress, eventRound: { $gt: 0 } } );
+  const totalVotes = accumulateBets(numOfResults, totalVoteRound);
+
+  // Add result set amount
+  totalVotes[resultIndex] += resultSetAmount
```
dwalintukan: this wont work. `totalVotes` is an array of strings. suggest to just add the resultSet obj to totalVoteRound before calling accumulateBets()

src/graphql/queries/total-result-bets.js
```diff
@@ -27,19 +27,40 @@ module.exports = async (
...
 
   // Accumulate all better bets
-  let betterBets;
+  let betterVotes;
   if (betterAddress) {
-    bets = await DBHelper.findBet({ eventAddress, betterAddress });
-    betterBets = accumulateBets(numOfResults, bets);
+    betterVoteRound = await DBHelper.findBet({ eventAddress, betterAddress, eventRound: { $gt: 0 } });
+    betterVotes = accumulateBets(numOfResults, betterVoteRound);
+    if (betterAddress === centralizedOracleAddress) betterVotes[resultIndex] += resultSetAmount;
```
dwalintukan: same here. need to add resultSet obj to betterVoteRound array before accum.

src/graphql/queries/total-result-bets.js
henryzbw: its the same as for the betting round, but roundBets can get other round

src/graphql/queries/total-result-bets.js
```diff
@@ -27,19 +29,42 @@ module.exports = async (
...
+  // Accumulate better round 0 bets
+  let betterBets
+  if (betterAddress) {
+    betterBetRound = await DBHelper.findBet({ eventAddress, eventRound: 0, betterAddress, txStatus: TX_STATUS.SUCCESS });
+    betterBets = accumulateBets(numOfResults, betterBetRound);
+  }
+
+  // Get result set
+  const resultSet = await DBHelper.findOneResultSet({ eventAddress, eventRound: 0, txStatus: TX_STATUS.SUCCESS });
+  const { centralizedOracleAddress } = resultSet;
```
dwalintukan: `resultSet` can be null if calling this during round 0.

PR #217, opened 2019-06-10
Sync Parser Tests

dwalintukan:


dwalintukan:
@BernardHan ya just the sync parsers

PR #221, opened 2019-06-11
Add previous round bets

henryzbw:
Used to show last round result, totalbets[resultIndex] and userbets[resultIndex]

### DIFF
--- a/src/graphql/resolvers.js
+++ b/src/graphql/resolvers.js
@@ -103,6 +103,8 @@ module.exports = {
           });
         }
 
+        const previousRoundUserBets = fill(Array(numOfResults), '0');
+        const previousRoundBets = fill(Array(numOfResults), '0');
         // Add result set amount if round 1
         if (currentRound === 1) {
           const resultSet = await DBHelper.findOneResultSet({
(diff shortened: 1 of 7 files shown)


dwalintukan:
The `roundBets` resolver is getting too convoluted with logic now. Let's just simplify this and return an array of all the amounts for each round.

```
type RoundBets {
  totalRoundBets: [String]
  userRoundBets: [String]
  singleTotalRoundBets: [[String]]
  singleUserRoundBets: [[String]]
}

singleTotalRoundBets: [
  [1,2,3], // index 0 = totals for round 0
  [4,5,6] // index 1 = totals for round 1
]

// same thing for singleUserRoundBets
```

henryzbw:
I refine them to return roundbets for each round 
```
type RoundBets {
  singleTotalRoundBets: [[String!]]
  singleUserRoundBets: [[String!]]
}
```
I removed 
```
  totalRoundBets: [String!]
  userRoundBets: [String!]
```
since it only returns current round bets, so its duplicate now

henryzbw:
Tried to fix the tests, but don't know why it fails. sometimes it can pass, but most of the time it fails

dwalintukan:
@henryzbw hm seems to be bugged, just remove the roundBets from the schema-helper.

src/graphql/resolvers.js
```diff
@@ -61,72 +61,48 @@ module.exports = {
...
-        // Fetch all bets for this round
-        const bets = await DBHelper.findBet({
+        // Fetch all bets for this event
+        let bets = await DBHelper.findBet({
           txStatus: TX_STATUS.SUCCESS,
           eventAddress: address,
-          eventRound: currentRound,
         });
 
+        const singleTotalRoundBets = []
```
dwalintukan: run linter

src/graphql/resolvers.js
```diff
@@ -61,72 +61,48 @@ module.exports = {
...
+          singleUserRoundBets.push(fill(Array(numOfResults), '0'));
+        }
+
+        const resultSet = await DBHelper.findOneResultSet({
+          txStatus: TX_STATUS.SUCCESS,
+          eventAddress: address,
+          eventRound: 0,
+        });
+
+        if (resultSet && resultSet.length != 0) {
```
dwalintukan: !==

PR #232, opened 2019-06-13
Biggest winner paginated

BernardHan:


src/graphql/queries/biggest-winners.js
```diff
@@ -64,7 +64,28 @@ module.exports = async (
     }));
   });
   await Promise.all(promises);
-  winnings = _orderBy(winnings, ['amount'], ['desc']);
+  let items = winnings;
```
dwalintukan: dont need to reassign, just do `{ items: winnings }` at the return

src/graphql/queries/biggest-winners.js
```diff
@@ -64,7 +64,28 @@ module.exports = async (
     }));
   });
   await Promise.all(promises);
-  winnings = _orderBy(winnings, ['amount'], ['desc']);
```
dwalintukan: why remove the ordering? youll have to implement `orderBy` param for the query then.

BernardHan: the orderBy gives weird order, one result was 7, 200, 0, don't know how.

So I'm just using array.sort.

dwalintukan: hm oh prob because its sorting by Strings instead of numbers

dwalintukan:
🚢

PR #241, opened 2019-06-18
Pending transactions

BernardHan:


src/graphql/schema.js
```diff
@@ -301,6 +301,7 @@ input WithdrawFilter {
 input TransactionFilter {
   eventAddress: String
   transactorAddress: String
+  txid: String
```
dwalintukan: create a new `PendingTransactionsFilter` since we dont use txid in the `Transactions` query.

src/graphql/queries/pending-transactions.js
```diff
@@ -0,0 +1,114 @@
...
+} = {}) => {
+  if (!eventAddress && !transactorAddress && !txid) {
+    throw Error('eventAddress or transactorAddress or txid missing in filters');
+  }
+
+  const filter = {};
+  if (eventAddress) filter.eventAddress = eventAddress;
+  if (transactorAddress) filter.transactorAddress = transactorAddress;
+  if (txid) filter.txid = txid;
+  filter.txStatus = TX_STATUS.PENDING;
```
dwalintukan: do we really only want to get the `PENDING` txs only? 

in our discussion, we talked about checking specific txids to see if they confirmed or not (`SUCCESS/FAIL`). the problem with checking for `PENDING` is say for some reason the pending tx didnt make it to the server, then you call `pendingTransactions` and don't see it on the list. here you are _assuming_ that the pending tx 1. made it to the server and 2. that it confirmed on the basis that its not on the `PENDING` list.

it would be more explicit to pass in an array of txids, return those txs and check the statuses of each one. otherwise i would say just alter the `transactions` query to add in `txid`. let me know what you think.

BernardHan: In our discussion, we did talk about using txid to check, but then I thought in mobile, JS is unable to get the txid because the flow stops. So I used this method, which also won't work in mobile because I forgot the pending tx never goes to server if made in mobile.

But if mobile issue is fixed, both ways work. This method does not assume the pending tx gets confirmed if the pending list is empty. Our UI doesn't really show the failed txs, so no matter it's confirmed or fail, JS would stop pulling data as long as the pending list is empty, which is equivalent to if the tx with the txid become `NOT PENDING`.

One extra benefit to use pending list, is that it's convenient to just put the pending list on top of confirmed list. Otherwise, UI needs to run extra loops on current transaction list to filter out`SUCCESS` pendings, and add  `PENDING` pendings back.

PR #242, opened 2019-06-19
Add withdraw list

BernardHan:
Think we need to migrate?

I wrote a migration script, but haven't tested yet, because I need to check if the logic in `db/index`is correct.

### DIFF
--- a/src/db/migrations/migration1.js
+++ b/src/db/migrations/migration1.js
@@ -1,0 +1,25 @@
+const logger = require('../../utils/logger');
+const DBHelper = require('../db-helper');
+const { TX_STATUS } = require('../../constants');
+
+async function migration1() {
+  try {
+    const events = await DBHelper.findEvent({ txStatus: TX_STATUS.WITHDRAW });
+    /* eslint-disable no-await-in-loop */
+    for (let i = 0; i < events.length; i++) {
+      const event = events[i];
+      const withdraws = await DBHelper.findWithdraw({ eventAddress: event.address });
+      const withdrawnList = [];
+      withdraws.map(withdraw => withdrawnList.push(withdraw.winnerAddress));
+      await DBHelper.updateEvent(
+        event.txid,
+        { withdrawnList },
+      );
+    }
+  } catch (err) {
+    logger.error(`Migration 1 error: ${err.message}`);
+    throw Error(`Migration 1 error: ${err.message}`);
+  }
+}
+
+module.exports = migration1;
--- a/src/graphql/mutations/add-pending-withdraw.js
+++ b/src/graphql/mutations/add-pending-withdraw.js
@@ -1,4 +1,4 @@
-const { isNull } = require('lodash');
+const { isNull, isUndefined } = require('lodash');
 const Withdraw = require('../../models/withdraw');
 const logger = require('../../utils/logger');
 const { getTransaction } = require('../../utils/web3-utils');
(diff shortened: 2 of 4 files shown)


src/db/migrations/migration1.js
```diff
@@ -0,0 +1,25 @@
+const logger = require('../../utils/logger');
+const DBHelper = require('../db-helper');
+const { TX_STATUS } = require('../../constants');
+
+async function migration1() {
+  try {
+    const events = await DBHelper.findEvent({ txStatus: TX_STATUS.WITHDRAW });
```
dwalintukan: cant we just loop through all the `successful` withdraws instead of having to fetch events first? you will have to alter the `updateEvent` func to be able to look it up by `address` as well.

src/graphql/mutations/add-pending-withdraw.js
```diff
@@ -22,6 +22,12 @@ module.exports = async (root, data) => {
   const txReceipt = await getTransaction(txid);
   await DBHelper.insertTransactionReceipt(txReceipt);
 
+  const event = await DBHelper.findOneEvent({ address: eventAddress });
```
dwalintukan: please add comment above line

src/graphql/mutations/add-pending-withdraw.js
```diff
@@ -22,6 +22,12 @@ module.exports = async (root, data) => {
...
+  if (isNull(event)) throw Error('Event does not exist');
+  if (isNull(event.withdrawnList) || isUndefined(event.withdrawnList)) event.withdrawnList = [];
```
dwalintukan: i made a helper func for this in `utils/index`

src/graphql/mutations/add-pending-withdraw.js
dwalintukan: please note that since we are adding to the list in the mutation (I assume you are doing this to show the pending status in the dashboard cards?), we need to handle if the withdraw tx fails when the tx gets confirmed. we should take it off the list in the event of a failed tx. see `sync/winnings-withdrawn.js L77`

dwalintukan: also this whole logic of adding it to the `withdrawnList` needs to be in the `sync/winnings-withdraw success` logic too. if we ever delete the db later and start from scratch, we want to populate the `withdrawnList` from all the withdraw events on chain.

dwalintukan:
@BernardHan oh also, can you fix the logic of the migrations. in `db/index L118`, that logic is actually not doing it in sequential order. async functions in loops execute parallel instead of sequentially which is not what we need in migrations. we need to ensure sequential order of each migration in case a future migration is dependent on a previous one.

this lib https://github.com/caolan/async should be able to do what you need.

src/db/migrations/migration1.js
BernardHan: so update a withdrawer one at a time, instead of updating an event withdrawnList as a whole in one time?

BernardHan: If I understand that correctly, why? I thought the fewer update the quicker?

dwalintukan: Oops i forgot to submit the update. How about this solution?
1. Fetch all successful withdraws
2. Create a map: eventAddress -> [] of withdraw address
3. Populate map as you iterate withdraw list
4. Iterate map keys and update events so they have the full withdrawn list
This should lower the number of calls to: `findWithdraws + (updateEvent * numEvents)` vs `findEvent + (findWithdraws * numEvents) + (update * numEvents)`

PR #246, opened 2019-06-19
Add withdrawn list

BernardHan:


src/db/index.js
```diff
@@ -113,27 +114,36 @@ async function applyMigrations() {
...
+  async.whilst(
+    condition => condition(null, i < migrations.length),
+    (callback) => {
+      const migration = migrations[i];
+      i++;
+      try {
+        if (Number(migration.number) === lastMigration + 1) {
+          // Run migration
+          logger.info(`Running migration ${migration.number}...`);
+          migration.migrate(db);
```
dwalintukan: need to await here? the migration func is async.

src/db/index.js
```diff
@@ -113,27 +114,36 @@ async function applyMigrations() {
...
-  }
-  /* eslint-enable no-restricted-syntax, no-await-in-loop */
-
-  logger.info('Migrations complete.');
+      callback(null, i);
+    },
+    (err, number) => {
+      if (err) {
+        logger.error(`Migration ${number} error: ${err.message}`);
+        return;
```
dwalintukan: throw here instead of return. we cant continue if a migration failed.

src/db/index.js
```diff
@@ -113,27 +114,36 @@ async function applyMigrations() {
...
+      try {
+        if (Number(migration.number) === lastMigration + 1) {
+          // Run migration
+          logger.info(`Running migration ${migration.number}...`);
+          migration.migrate(db);
+          logger.info(`Migration ${migration.number} done`);
+          // Track the last migration number
+          lastMigration = migration.number;
+          fs.outputFileSync(migrationTrackPath, `LAST_MIGRATION=${migration.number}\n`);
+          logger.info('Write file done');
```
dwalintukan: nit: more descriptive log, "migration.dat updated"?

src/db/index.js
```diff
@@ -113,27 +114,36 @@ async function applyMigrations() {
...
+        }
+      } catch (err) {
+        callback(err, migration.number);
```
dwalintukan: The loop will stop here cause the `err != null` right?

src/db/migrations/migration1.js
```diff
@@ -0,0 +1,31 @@
...
+      const withdraw = withdraws[i];
+      let withdrawnList = [];
+      if (withdrawnLists.has(withdraw.eventAddress)) {
+        withdrawnList = withdrawnLists.get(withdraw.eventAddress);
+      }
+      withdrawnList.push(withdraw.winnerAddress);
+      withdrawnLists.set(withdraw.eventAddress, withdrawnList);
+    }
+
+    withdrawnLists.forEach((value, key) => {
```
dwalintukan: Need to create a list of promises, add all the update promises, then call `await Promise.all`. This current logic wont wait for all the db updates to be done before continuing.

src/db/migrations/migration1.js
```diff
@@ -0,0 +1,31 @@
...
+      DBHelper.updateEventByAddress(
```
dwalintukan: await

src/db/db-helper.js
```diff
@@ -206,6 +216,24 @@ module.exports = class DBHelper {
     }
   }
 
+  static async updateEventWithdrawnList(withdraw) {
+    try {
+      const event = await DBHelper.findOneEvent({ address: withdraw.eventAddress });
+      if (isNull(event)) throw Error('Event does not exist');
```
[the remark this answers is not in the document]
BernardHan: The func would return at the throw I think?

src/db/index.js
BernardHan: Looks like it doesn't allow await, only callback

src/db/index.js
BernardHan: right

BernardHan:
@dwalintukan I have rewrote the apply migration, and made the changes

I made detailed comments to explain what's happening.

PR #250, opened 2019-06-21
Graphql test

BernardHan:
Only have tests for `add-pending-event` now, if it looks good, I will keep working on the rest.

### DIFF
--- a/test/graphql/mutations.tests.js
+++ b/test/graphql/mutations.tests.js
@@ -12,7 +12,7 @@ const tester = new EasyGraphQLTester(schema);
 
 describe('graphql/mutations', () => {
   describe('add-pending-event', () => {
-    it('General test', async () => {
+    it('Pass all arguments', async () => {
       const valid = `
         mutation {
           addPendingEvent(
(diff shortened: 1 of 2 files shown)


test/graphql/mutations.tests.js
```diff
@@ -0,0 +1,516 @@
...
+      const valid = `
+        mutation {
+          addPendingEvent(
+            txid: "0xd5d087daabc73fc6cc5d9c1131b93acbd53a2428"
+            ownerAddress: "0xd5d087daabc73fc6cc5d9c1131b93acbd53a2428"
+            name: "General test"
+            results: ["1", "2"]
+            numOfResults: 2
+            centralizedOracle: "0xd5d087daabc73fc6cc5d9c1131b93acbd53a2428"
+            betStartTime: 1560965704
```
dwalintukan: remove from all tests since its no longer part of the mutation schema

test/graphql/mutations.tests.js
```diff
@@ -0,0 +1,516 @@
...
+            betEndTime: 1560965704
+            resultSetStartTime: 1560965704
+            resultSetEndTime: 1560965704
```
dwalintukan: remove from all tests

test/graphql/mutations.tests.js
```diff
@@ -0,0 +1,516 @@
...
+  PAGINATED_RESULT_SETS,
+  PAGINATED_WITHDRAWS,
+} = require('./schema-helper');
+const schema = require('../../src/graphql/schema');
+
+const tester = new EasyGraphQLTester(schema);
+
+describe('graphql/mutations', () => {
+  describe('add-pending-event', () => {
+    it('Pass all arguments', async () => {
```
dwalintukan: nit: write the test name like you would read it as a sentence: `it passes with all arguments`

test/graphql/mutations.tests.js
```diff
@@ -0,0 +1,516 @@
...
+            language: "en-US"
+          ) {
+            ${MULTIPLE_RESULTS_EVENT}
+          }
+        }
+      `;
+      tester.test(false, valid, {});
+    });
+
+    it('Miss betStartTime', async () => {
```
dwalintukan: this shouldnt fails since we removed this field from the mutation

test/graphql/mutations.tests.js
```diff
@@ -0,0 +1,516 @@
...
+            language: "en-US"
+          ) {
+            ${MULTIPLE_RESULTS_EVENT}
+          }
+        }
+      `;
+      tester.test(false, valid, {});
+    });
+
+    it('Miss resultSetEndTime', async () => {
```
dwalintukan: this shouldnt fail since we removed the field from the mutation

test/graphql/mutations.tests.js
```diff
@@ -0,0 +1,516 @@
...
+            resultSetEndTime: 1560965704
+          ) {
+            ${MULTIPLE_RESULTS_EVENT}
+          }
+        }
+      `;
+      tester.test(false, valid, {});
+    });
+
+    it('Txid should be string', async () => {
```
dwalintukan: nit: `it should fail is txid is not a string`

test/graphql/mutations.tests.js
```diff
@@ -0,0 +1,516 @@
...
+      const valid = `
+        mutation {
+          addPendingEvent(
+            txid: 12
```
dwalintukan: also add another test case if `txid: null`. you can add it in this same `it` block since it covers it.

test/graphql/mutations.tests.js
```diff
@@ -0,0 +1,516 @@
...
+      `;
+      tester.test(false, valid, {});
+    });
+
+    it('ownerAddress should be string', async () => {
+      const valid = `
+        mutation {
+          addPendingEvent(
+            txid: "0xd5d087daabc73fc6cc5d9c1131b93acbd53a2428"
+            ownerAddress: null
```
dwalintukan: add another case in here if `ownerAddress` is a number. when writing validation tests, think of all the different kinds of types that can potentially be passed in. even passing in an `object` or `array` can be a case,

test/graphql/mutations.tests.js
```diff
@@ -0,0 +1,516 @@
...
+            language: "en-US"
+          ) {
+            ${MULTIPLE_RESULTS_EVENT}
+          }
+        }
+      `;
+      tester.test(false, valid, {});
+    });
+
+    it('betStartTime should be int', async () => {
```
dwalintukan: remove test, mutation doesnt use this field anymore

test/graphql/mutations.tests.js
```diff
@@ -0,0 +1,2249 @@
...
+          ) {
+            ${MULTIPLE_RESULTS_EVENT}
+          }
+        }
+      `;
+      tester.test(true, valid, {});
+    });
+
+    it('It should fail if txid missed', async () => {
+      const valid = `
```
dwalintukan: maybe name `valid` to something like `mutation` so you dont need to change `valid` -> `invalid` (and vice versa) for tests that fail?

PR #257, opened 2019-06-26
Db tests

henryzbw:


test/db/index.js
```diff
@@ -0,0 +1,86 @@
...
+        await DBHelper.insertBlock(toBeInsertedBlockNum, toBeInsertedBlockTime);
+        blocks = await db.Blocks.find({});
+        blocks.length.should.equal(1);
+        blocks[0].blockNum.should.equal(toBeInsertedBlockNum);
+        blocks[0].blockTime.should.equal(toBeInsertedBlockTime);
+      });
+
+      it('should not insert with existed block num', async () => {
+        const toBeInsertedFailBlockTime = Date.now();
+        await DBHelper.insertBlock(toBeInsertedBlockNum, toBeInsertedFailBlockTime);
```
dwalintukan: this should throw here so you need to catch the throw and assert there is an error

test/db/index.js
```diff
@@ -0,0 +1,86 @@
...
+        latestBlock = await DBHelper.findLatestBlock();
+        latestBlock[0].blockNum.should.equal(latestBlockNum);
+        latestBlock[0].blockTime.should.equal(latestBlockTime);
+      });
+    });
+
+    describe('find block', () => {
+      const toBeInsertedBlockNum = 10001;
+      const toBeInsertedBlockTime = Date.now();
+      it('find exitsed block in block db', async () => {
```
dwalintukan: typo

test/db/index.js
```diff
@@ -0,0 +1,86 @@
...
+        await DBHelper.insertBlock(toBeInsertedBlockNum, toBeInsertedBlockTime);
+        const block = await DBHelper.findOneBlock({ blockNum: toBeInsertedBlockNum });
+        should.exist(block);
+        block.blockNum.should.equal(toBeInsertedBlockNum);
+        block.blockTime.should.equal(toBeInsertedBlockTime);
+      });
+
+      it('find non-exitsed block in block db', async () => {
```
dwalintukan: typo

test/db/index.js
```diff
@@ -0,0 +1,86 @@
...
+        const block = await DBHelper.findOneBlock({ blockNum: -1 });
```
dwalintukan: also add a test to use a positive number here. some random block # not in db.

test/db/index.js
henryzbw: but in the DBHelper.insertBlock, it will check if it is existed, it it's already there, the method just return, no error throwed in that method.

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+      let blockCount;
+      it('find latest on empty block db', async () => {
+        db.Blocks.remove({}, { multi: true });
+        blockCount = await db.Blocks.count({});
+        blockCount.should.equal(0);
+        const latestBlock = await DBHelper.findLatestBlock();
+        latestBlock.length.should.equal(0);
+      });
+
+      it('find latest on non-empty block db', async () => {
```
dwalintukan: `should return the latest block with a non-empty db`

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+        blocks.length.should.equal(1);
+        blocks[0].blockNum.should.equal(toBeInsertedBlockNum);
+        blocks[0].blockTime.should.equal(toBeInsertedBlockTime);
+      });
+    });
+
+    describe('find latest block', () => {
+      let blockCount;
+      it('find latest on empty block db', async () => {
+        db.Blocks.remove({}, { multi: true });
```
dwalintukan: dont need to change it here but for future reference. if you are doing tests which require having a constant state (db in these cases), a good practice is to optimize the test order itself.

for example, for the Blocks db, you start with all your tests which require an empty db first. then you can do all the tests which check for existing entries in the db after you add them in. that way you don't have to keep clearing the db for the empty tests. it's like following a lifecycle for the test db.

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+        latestBlock = await DBHelper.findLatestBlock();
+        latestBlock[0].blockNum.should.equal(latestBlockNum);
+        latestBlock[0].blockTime.should.equal(latestBlockTime);
+      });
+    });
+
+    describe('find block', () => {
+      const toBeInsertedBlockNum = 10001;
+      const toBeInsertedBlockTime = Date.now();
+      it('find existed block in block db', async () => {
```
dwalintukan: finds the existing block by blockNum

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+        await DBHelper.insertBlock(toBeInsertedBlockNum, toBeInsertedBlockTime);
+        const block = await DBHelper.findOneBlock({ blockNum: toBeInsertedBlockNum });
+        should.exist(block);
+        block.blockNum.should.equal(toBeInsertedBlockNum);
+        block.blockTime.should.equal(toBeInsertedBlockTime);
+      });
+
+      it('find non-existed block in block db', async () => {
```
dwalintukan: does not find the non-existing block by blockNum

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+      blockNum: 4400858,
+      transactionHash: '0x81fcd2aec9f2f45a96eea6deb858a76fecce619df091f62ec92ec3bce6181d18',
+      from: '0x47ba776b3ed5d514d3e206ffee72fa483baffa7e',
+      to: '0x809388c8770c578cb87df1d1e84e3436d8156fda',
+      contractAddress: null,
+      cumulativeGasUsed: 207559,
+      gasUsed: 207559,
+      gasPrice: '30000000',
+    };
+    describe('insert', () => {
```
dwalintukan: 👾 moar whitespace!

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+      });
+    });
+  });
+
+
+  /* TransactionReceipts */
+  describe('test transactionReceipts db', () => {
+    let txReceiptsCount;
+    let txReceipt;
+    const toBeInsertedTxReceipt = {
```
dwalintukan: better naming would be: `UnconfirmedTxReceipt`

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+      blockHash: '0x0000000000000000000000000000000000000000000000000000000000000000',
+      blockNum: null,
+      transactionHash: '0x81fcd2aec9f2f45a96eea6deb858a76fecce619df091f62ec92ec3bce6181d18',
+      from: '0x47ba776b3ed5d514d3e206ffee72fa483baffa7e',
+      to: '0x809388c8770c578cb87df1d1e84e3436d8156fda',
+      cumulativeGasUsed: 500000,
+      gasUsed: 500000,
+      gasPrice: '47619047620',
+    };
+    const toBeInsertedExistedTxReceipt = {
```
dwalintukan: `confirmedTxReceipt`

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+        await DBHelper.insertTransactionReceipt(toBeInsertedExistedTxReceipt);
+        txReceipt = await db.TransactionReceipts.find({});
+        txReceipt.length.should.equal(1);
+        expect(txReceipt[0]).excluding(["_id", "gasPrice"]).to.deep.equal(toBeInsertedExistedTxReceipt)
+        txReceipt[0].gasPrice.should.equal(toBeInsertedTxReceipt.gasPrice);
+      });
+    });
+
+    describe('find one transaction receipt', () => {
+      it('should find existed transaction Hash', async () => {
```
dwalintukan: should find existing tx receipt by transactionHash

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+        txReceipt = await DBHelper.findOneTransactionReceipt({transactionHash: toBeInsertedExistedTxReceipt.transactionHash});
+        should.exist(txReceipt);
+        expect(txReceipt).excluding(["_id", "gasPrice"]).to.deep.equal(toBeInsertedExistedTxReceipt)
+        txReceipt.gasPrice.should.equal(toBeInsertedTxReceipt.gasPrice);
+      })
+
+      it('should not find non-existed transaction Hash', async () => {
```
dwalintukan: should not find non-existing tx receipt by transactionHash

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+        txReceipt = await DBHelper.findOneTransactionReceipt({transactionHash: 123321});
+        should.not.exist(txReceipt)
+      })
+    })
+
+    describe('update transaction receipt', () => {
+      it('should update existed transaction Hash transaction receipt', async () => {
```
dwalintukan: should update existing tx receipt

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+          gasUsed: 324432,
+          gasPrice: '60000000',
+        };
+        await DBHelper.updateTransactionReceipt(toBeUpdatedExistedTxReceipt);
+        txReceipt = await DBHelper.findOneTransactionReceipt({transactionHash: toBeUpdatedExistedTxReceipt.transactionHash});
+        expect(txReceipt).excluding("_id").to.deep.equal(toBeUpdatedExistedTxReceipt)
+      })
+    })
+
+    it('should fail update non-existed transaction Hash transaction receipt', async () => {
```
dwalintukan: should throw trying to update a non-existing tx receipt

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+      await DBHelper.updateTransactionReceipt(toBeUpdatedExistedTxReceipt);
+      txReceipt = await DBHelper.findOneTransactionReceipt({transactionHash: toBeUpdatedExistedTxReceipt.transactionHash});
+      should.not.exist(txReceipt);
+    })
+  });
+
+  /* Events */
+  describe('test event db', () => {
+    const mockEvent = [
+      {"txType":"CREATE_EVENT","txid":"0xa47a223155ee3f9e1d313735547e7ad3e299d20992c0524f6b9694984100f7b0","txStatus":"SUCCESS","blockNum":4242848,"address":"0x60f243ed4a99e3309ef67d9486b9db14334bc6d4","ownerAddress":"0xd5d087daabc73fc6cc5d9c1131b93acbd53a2428","version":6,"name":"Test v6","results":["Invalid","1","2"],"numOfResults":3,"centralizedOracle":"0xd5d087daabc73fc6cc5d9c1131b93acbd53a2428","betStartTime":1561082467,"betEndTime":1561168846,"resultSetStartTime":1561168846,"resultSetEndTime":1561341646,"escrowAmount":"100000000","arbitrationLength":300,"thresholdPercentIncrease":"10","arbitrationRewardPercentage":10,"currentRound":0,"currentResultIndex":255,"consensusThreshold":"1000000000","arbitrationEndTime":0,"status":"OPEN_RESULT_SETTING","language":"zh-Hans-CN"},
```
dwalintukan: please format all these so its readable. can try using a pretty printer

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+      {"txType":"CREATE_EVENT","txid":"0x050497d3ed87c02246ba45ccd965f28e4194aace1167a25e46a2f6e29407d01c","txStatus":"SUCCESS","blockNum":3778535,"address":"0x80b35153213219ac3a3a3f28fe3cd3996518804a","ownerAddress":"0x939592864c0bd3355b2d54e4fa2203e8343b6d6a","version":4,"name":"test","results":["Invalid","1","2"],"numOfResults":3,"centralizedOracle":"0x939592864c0bd3355b2d54e4fa2203e8343b6d6a","betStartTime":1559689800,"betEndTime":1559691600,"resultSetStartTime":1559691600,"resultSetEndTime":1559863179,"escrowAmount":"100000000","arbitrationLength":86400,"thresholdPercentIncrease":"10","arbitrationRewardPercentage":10,"currentRound":0,"currentResultIndex":255,"consensusThreshold":"10000000000","arbitrationEndTime":0,"status":"OPEN_RESULT_SETTING","language":"zh-Hans-CN"}
+    ]
+    let eventCount;
+    describe('find event', () => {
+      it('find existed event', async () => {
```
dwalintukan: should return existing events if passing sort

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+      {"txType":"CREATE_EVENT","txid":"0x050497d3ed87c02246ba45ccd965f28e4194aace1167a25e46a2f6e29407d01c","txStatus":"SUCCESS","blockNum":3778535,"address":"0x80b35153213219ac3a3a3f28fe3cd3996518804a","ownerAddress":"0x939592864c0bd3355b2d54e4fa2203e8343b6d6a","version":4,"name":"test","results":["Invalid","1","2"],"numOfResults":3,"centralizedOracle":"0x939592864c0bd3355b2d54e4fa2203e8343b6d6a","betStartTime":1559689800,"betEndTime":1559691600,"resultSetStartTime":1559691600,"resultSetEndTime":1559863179,"escrowAmount":"100000000","arbitrationLength":86400,"thresholdPercentIncrease":"10","arbitrationRewardPercentage":10,"currentRound":0,"currentResultIndex":255,"consensusThreshold":"10000000000","arbitrationEndTime":0,"status":"OPEN_RESULT_SETTING","language":"zh-Hans-CN"}
+    ]
+    let eventCount;
+    describe('find event', () => {
```
dwalintukan: also add tests if you dont pass in a `sort`

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+        await DBHelper.insertEvent(mockEvent[2]);
+        eventCount = await db.Events.count({});
+        eventCount.should.equal(3);
+        const event = await DBHelper.findOneEvent({txid: mockEvent[2].txid});
+        should.exist(event);
+        expect(event).excluding("_id").to.deep.equal(mockEvent[2]);
+      })
+
+      it('insert existed txid event', async () => {
+        const newEventWithExistedTxid = {"txType":"CREATE_EVENT","txid":"0x050497d3ed87c02246ba45ccd965f28e4194aace1167a25e46a2f6e29407d01c","txStatus":"FAIL","blockNum":3728535,"address":"0x80b351543219ac3a3a3f28fe3cd3996518804a","ownerAddress":"0x939592432410bd3355b2d54e4fa2203e8343b6d6a","version":5,"name":"N","results":["Invalid","12","2"],"numOfResults":3,"centralizedOracle":"0x939592864c0b32d3355b2d15a2203e8343b6d6a","betStartTime":1559689700,"betEndTime":1559692600,"resultSetStartTime":1559692600,"resultSetEndTime":1559864179,"escrowAmount":"110000000","arbitrationLength":86400,"thresholdPercentIncrease":"10","arbitrationRewardPercentage":10,"currentRound":0,"currentResultIndex":255,"consensusThreshold":"10000000000","arbitrationEndTime":0,"status":"OPEN_RESULT_SETTING","language":"en-US"}
```
dwalintukan: pretty print

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+        const event = await DBHelper.findOneEvent({txid: newEventWithExistedTxid.txid});
+        should.exist(event);
+        expect(event).excluding(["_id", "language"]).to.deep.equal(newEventWithExistedTxid);
+        event.language.should.equal(mockEvent[2].language);
+      })
+    })
+
+    describe('update event', () => {
+      it('update event by txid', async () => {
+        const updateNewEvent = {"txType":"CREATE_EVENT","txid":"0x050497d3ed87c02246ba45ccd965f28e4194aace1167a25e46a2f6e29407d01c","txStatus":"SUCCESS","blockNum":123442,"address":"0x382737120ad2323","ownerAddress":"0xea5854445251393b0e6f0a870799da15c003fc03","version":6,"name":"N","results":["Invalid","12","2"],"numOfResults":3,"centralizedOracle":"0x939592864c0b32d3355b2d15a2203e8343b6d6a","betStartTime":1559689700,"betEndTime":1559692600,"resultSetStartTime":1559692600,"resultSetEndTime":1559864179,"escrowAmount":"110000000","arbitrationLength":86400,"thresholdPercentIncrease":"10","arbitrationRewardPercentage":10,"currentRound":0,"currentResultIndex":255,"consensusThreshold":"10000000000","arbitrationEndTime":0,"status":"OPEN_RESULT_SETTING","language":"en-US"}
```
dwalintukan: use a real length address

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+        eventCount = await db.Events.count({});
+        eventCount.should.equal(3);
+        event = await DBHelper.findOneEvent({txid: mockEvent[2].txid});
+        should.exist(event);
+        expect(event).excluding(["_id"]).to.deep.equal(mockEvent[2]);
+      })
+    })
+
+    describe('event status update', () => {
+      let createdEvent = {"txType":"CREATE_EVENT","txid":"0x9628812210d8e7d071ee343ece4d79c248191fdf76938cecd572158c475cb7f5","txStatus":"SUCCESS","blockNum":4404133,"address":"0x0c97ba43149c6eee920ef1e9d410316905c1ce8a","ownerAddress":"0x47ba776b3ed5d514d3e206ffee72fa483baffa7e","version":6,"name":"TEST EVENT STATUS","results":["Invalid","1","2"],"numOfResults":3,"centralizedOracle":"0x47ba776b3ed5d514d3e206ffee72fa483baffa7e","betStartTime":1561585230,"betEndTime":1561671614,"resultSetStartTime":1561671714,"resultSetEndTime":1561844414,"escrowAmount":"100000000","arbitrationLength":86400,"thresholdPercentIncrease":"10","arbitrationRewardPercentage":10,"currentRound":0,"currentResultIndex":255,"consensusThreshold":"10000000000","arbitrationEndTime":0,"status":"CREATED","language":"en-US","withdrawnList":[]}
```
dwalintukan: pretty print

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+      it('should update event status to PRE_BETTING', async () => {
+        await DBHelper.insertEvent(createdEvent);
+        await DBHelper.updateEventStatusPreBetting(createdEvent.betStartTime - 1);
+        event = await DBHelper.findEvent({address: createdEvent.address});
+        event.length.should.equal(1);
+        expect(event[0]).excluding(["_id", 'status']).to.deep.equal(createdEvent);
+        event[0].status.should.equal(EVENT_STATUS.PRE_BETTING)
+      })
+
+      it('should update event status to BETTING', async () => {
```
dwalintukan: also check the upper bound as well `betEndTime - 1`

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+        await DBHelper.updateEventStatusBetting(createdEvent.betStartTime + 1);
+        event = await DBHelper.findEvent({address: createdEvent.address});
+        event.length.should.equal(1);
+        expect(event[0]).excluding(["_id", 'status']).to.deep.equal(createdEvent);
+        event[0].status.should.equal(EVENT_STATUS.BETTING)
+      })
+
+      it('should update event status to PRE_RESULT_SETTING', async () => {
```
dwalintukan: check lower bound, `betEndTime + 1`

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+        await DBHelper.updateEventStatusPreResultSetting(createdEvent.resultSetStartTime - 1);
+        event = await DBHelper.findEvent({address: createdEvent.address});
+        event.length.should.equal(1);
+        expect(event[0]).excluding(["_id", 'status']).to.deep.equal(createdEvent);
+        event[0].status.should.equal(EVENT_STATUS.PRE_RESULT_SETTING)
+      })
+
+      it('should update event status to ORACLE_RESULT_SETTING', async () => {
```
dwalintukan: check upper bound

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+
+      it('should update event status to WITHDRAWING', async () => {
+        await DBHelper.updateEventStatusWithdrawing(createdEvent.arbitrationEndTime);
+        event = await DBHelper.findEvent({address: createdEvent.address});
+        event.length.should.equal(1);
+        expect(event[0]).excluding(["_id", 'status']).to.deep.equal(createdEvent);
+        event[0].status.should.equal(EVENT_STATUS.WITHDRAWING)
+      })
+
+      describe('test event withdrawn list', () => {
```
dwalintukan: nit: inconsistent that testing withdrawn list has its own describe block but the other update statuses dont

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+        const withdraw = {
+          eventAddress: "0xfef4a675dba91b91608dc75c40deaca0333af514",
+          winnerAddress: "0x47ba776b3ed5d514d3e206ffee72fa483baffa7e",
+        }
+        it('should update event withdrawnlist', async () => {
+          await DBHelper.updateEventWithdrawnList(withdraw);
+          event = await DBHelper.findOneEvent({address: withdraw.eventAddress});
+          event.withdrawnList[0].should.equal(withdraw.winnerAddress);
+        })
```
dwalintukan: missing updateEventWithdrawnList tests:
1. does not add to withdrawnList if already on the list

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+        await DBHelper.updateEventByAddress("0x382737120ad2323", mockEvent[2]);
+        eventCount = await db.Events.count({});
+        eventCount.should.equal(3);
+        event = await DBHelper.findOneEvent({txid: mockEvent[2].txid});
+        should.exist(event);
+        expect(event).excluding(["_id"]).to.deep.equal(mockEvent[2]);
+      })
+    })
+
+    describe('event status update', () => {
```
dwalintukan: please go through the event status tests and add the missing ones. for each of the status change tests you need to test every corner case.

e.g.
updateEventStatusPreBetting(). these are all the tests you should cover:
1. it updates the event to pre-betting status (success case, all fields valid)
2. it does not update events that do not have TX_STATUS.SUCCESS (fail case, where all the other fields are valid except for the txStatus)
3. it does not update events that are already in EVENT_STATUS.PRE_BETTING (fail case, where all other fields are valid except for status)
4. it does not update if currBlockTime = betStartTime (fail case, where all other fields are valid except betStartTime)
4. it does not update events that currentRound > 0 (fail case, where all other fields are valid except currentRound)

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+        })
+      })
+    })
+
+  })
+
+  /* Bets */
+  describe('test bets db', () => {
+    const mockBets = [
+      {"txType":"BET","txid":"0x445ffdcfb4b3e30c3c8d73a7583230b60eff51054035389df7cf187f5d3f75b6","txStatus":"SUCCESS","blockNum":4154205,"eventAddress":"0xbeaa4358aa4434227c4545544e50c1b1a52a51c6","betterAddress":"0x939592864c0bd3355b2d54e4fa2203e8343b6d6a","resultIndex":2,"amount":"1100000000","eventRound":0},
```
dwalintukan: pretty print

test/db/index.js
```diff
@@ -0,0 +1,449 @@
...
+        bet = await DBHelper.findOneBet({txid: mockBets[2].txid});
+        should.exist(bet);
+        expect(bet).excluding("_id").to.deep.equal(mockBets[2]);
+      })
+
+      it('insert existed txid bet', async () => {
+        const newBetWithExistedTxid =       {"txType":"BET","txid":"0xccd0af2cb0299880101ff6558c3000a719e485677a9792bc24c4366ac78d1bf4","txStatus":"SUCCESS","blockNum":1233242,"eventAddress":"0xeaf92b5aabe6028fdbe5889a6705f322e2da8d","betterAddress":"0x47ba776b3ed5d5325345fee72fa483baffa7e","resultIndex":1,"amount":"300000000","eventRound":0}
+        await DBHelper.insertBet(newBetWithExistedTxid);
+        betCount = await db.Bets.count({});
+        betCount.should.equal(3);
```
dwalintukan: you should also do this check before `insertBet` is called so you know that the count didnt increase

test/db/index.js
henryzbw: it doesn't throw error. so maybe leave it as should fail?

dwalintukan: gotcha, fail is fine

test/db/index.js
```diff
@@ -0,0 +1,1539 @@
...
+          consensusThreshold: '10000000000',
+          arbitrationEndTime: 0,
+          status: 'CREATED',
+          language: 'en-US',
+          withdrawnList: [
+          ],
+        };
+        it('should not update event status to PRE_BETTING if txStatus is FAIL', async () => {
+          const createdEvent = {};
+          Object.assign(createdEvent, mockCreatedStatusEvent, { txStatus: TX_STATUS.FAIL, address: padHex(40), txid: padHex(64) });
```
dwalintukan: a failed tx wont have an address

test/db/index.js
```diff
@@ -0,0 +1,1539 @@
...
+          await DBHelper.insertEvent(createdEvent);
+          await DBHelper.updateEventStatusPreBetting(createdEvent.betStartTime - 1);
+          event = await DBHelper.findEvent({ address: createdEvent.address });
+          event.length.should.equal(1);
+          expect(event[0]).excluding(['_id']).to.deep.equal(createdEvent);
```
dwalintukan: dont need to check every field for this test. you wanna check the event `status` and make sure it stays `CREATED`.

instead of checking every field, check `txid` or `address` to verify that it is the event you inserted. and also check the `status` did not change from `CREATED`.

i know your check does handle it, but you want to be focused when testing a specific case. like for example, if some change broke another part of the event, say like the name or something. then this test would fail cause of that as well. yes it does make test writing more tedious, but it serves a purpose.

test/db/index.js
```diff
@@ -0,0 +1,1539 @@
...
+        });
+
+        it('should not update event status to PRE_BETTING if txStatus is PENDING', async () => {
+          const createdEvent = {};
+          Object.assign(createdEvent, mockCreatedStatusEvent, { txStatus: TX_STATUS.PENDING, address: padHex(40), txid: padHex(64) });
```
dwalintukan: no address for pending event

test/db/index.js
```diff
@@ -0,0 +1,1539 @@
...
+          await DBHelper.insertEvent(createdEvent);
+          await DBHelper.updateEventStatusPreBetting(createdEvent.betStartTime - 1);
+          event = await DBHelper.findEvent({ address: createdEvent.address });
+          event.length.should.equal(1);
+          expect(event[0]).excluding(['_id']).to.deep.equal(createdEvent);
```
dwalintukan: verify event & check status only

test/db/index.js
dwalintukan: please go through the all the event status tests and adjust accordingly

test/db/index.js
```diff
@@ -0,0 +1,1777 @@
...
+          language: 'en-US',
+          withdrawnList: [
+          ],
+        };
+
+        it('should not update event status to PRE_BETTING if txStatus is FAIL', async () => {
+          const createdEvent = {};
+          Object.assign(createdEvent, mockCreatedStatusEvent, { txStatus: TX_STATUS.FAIL, txid: padHex(64) });
+          // FAIL tx doesn't have address
+          delete createdEvent.address;
```
dwalintukan: cant you just do `{ address: undefined }` in `Object.assign`?

henryzbw: think its better to keep it as the real db which doesn't have the key

PR #263, opened 2019-06-26
Api test

BernardHan:


dwalintukan:
Please use `sinon` to do the mocking. Can check `multiple-results-event.tests.js` for small examples but referring to their docs is best.

See if you can stub the `contract.methods.calculateWinnings(address).call();`. Might be possible to stub the `contract.methods.xxx` functions. That would be the proper way so we can actually call `calculateWinnings()`.

Don't waste too much time on it though. Give it a go and see if possible, if not just do it like how you are doing it.

BernardHan:
@dwalintukan 

I found a way, which uses `callsFake` to replace `getContract`, then I mock the contract to have all the mocked methods.

I have to make `getContract` a method of `multiple-results-event.js`.

Please let me know if this is proper. If I can keeping doing this, I will complete the mocked contract and separate it out into `mock` folder.

src/api/multiple-results-event.js
```diff
@@ -13,13 +13,21 @@ const getContract = async (eventAddress) => {
 };
 
 module.exports = {
+  async getContract(eventAddress) {
```
dwalintukan: just add the func reference so you dont have to copy everything over.
```
module.exports = {
  getContract,
  async calculateWinnings(args) {},
  ...,
}
```

test/api/multiple-results-event.test.js
```diff
@@ -0,0 +1,133 @@
...
+describe('api/multiple-results-event', () => {
+  const contractAddress = 'a5b27c03e76d4cf10928120439fa96181f07520c';
+  let stubGetContract;
+
+  beforeEach(() => {
+      stubGetContract = sinon.stub(Event, 'getContract').callsFake((eventAddress) => {
+        if (eventAddress !== '0x8b2b1838efff78e5ed9e00d95d9ef071f2c27be6') return undefined;
+        return {
+          methods: {
+            calculateWinnings: (address) => {
```
dwalintukan: lets stub each contract method as well.
https://stackoverflow.com/a/44154342/2917742

Code not tested, just as example.
```
const stubCalculateWinnings = sinon.stub().returns('10200000000');  
sinon.stub().returns({
  methods: {
    calculateWinnings: stubCalculateWinnings,
    ...
  }
}
```

We aren't concerned about passing in the correct address or validating the contract instance, but more about getting the mock data from the API.

BernardHan: Think we should test if address is not listed, then no winnings?

dwalintukan: No need to test if the address is not listed. That is a test for the contract level. This test is more for the formatting of the data from the contract.

src/api/config-manager.js
```diff
@@ -12,9 +12,11 @@ const getContract = () => {
 };
 
 module.exports = {
+  getContract,
+
   async bodhiTokenAddress() {
     try {
-      return getContract().methods.bodhiTokenAddress().call();
+      return this.getContract().methods.bodhiTokenAddress().call();
```
dwalintukan: do you need to add `this.` to all of them? dont they reference the same function if you dont?

test/api/mock/contract.js
```diff
@@ -0,0 +1,87 @@
...
+      startingConsensusThreshold: stubStartingConsensusThreshold,
+      thresholdPercentIncrease: stubThresholdPercentIncrease,
+      isWhitelisted: stubIsWhitelisted,
+    }
+  }
+};
+
+module.exports = {
+  getContract,
+};
```
dwalintukan: add EOL

test/api/config-manager.test.js
```diff
@@ -0,0 +1,103 @@
...
+
+  afterEach(() => {
+    stubGetContract.restore();
+  });
+
+  describe('bodhiTokenAddress()', () => {
+    it('It returns the bodhiTokenAddress', async () => {
+      const res = await ConfigManager.bodhiTokenAddress();
+      assert.isDefined(res);
+      assert.isTrue(isString(res));
```
dwalintukan: fyi: chai has an `assert.isString`

test/api/config-manager.test.js
```diff
@@ -0,0 +1,103 @@
...
+    });
+  });
+
+  describe('eventEscrowAmount()', () => {
+    it('It returns the eventEscrowAmount', async () => {
+      const res = await ConfigManager.eventEscrowAmount();
+      assert.isDefined(res);
+      assert.isTrue(isString(res));
+      assert.isFalse(isNaN(res));
```
dwalintukan: nit: this test technically not needed since the isString test will ensure that this will never be true.

test/api/config-manager.test.js
```diff
@@ -0,0 +1,103 @@
...
+      const res = await ConfigManager.isWhitelisted({ address });
+      assert.isDefined(res);
+      assert.isTrue(isBoolean(res));
+    });
+
+    it('It throws if address is undefined', () => {
+      expect(ConfigManager.isWhitelisted()).to.be.rejectedWith(Error);
+    });
+  });
+});
```
dwalintukan: add EOL

test/api/multiple-results-event.test.js
```diff
@@ -0,0 +1,217 @@
...
+      const res = await Event.didWithdrawEscrow({ eventAddress, address });
+      assert.isDefined(res);
+      assert.isTrue(isBoolean(res));
+    });
+
+    it('It throws if eventAddress is undefined', () => {
+      expect(Event.didWithdrawEscrow({})).to.be.rejectedWith(Error);
+    });
+  });
+});
```
dwalintukan: add EOL

src/api/config-manager.js
BernardHan: Looks like if without `this.`, the test won't catch the stub version.

Issue #267, opened 2019-07-02
Make leaderboard data tables, store calculatewinnings into tables


BernardHan:
- [x] Create eventLeaderboard and globalLeaderboard tables and models

- [x] Make db helper methods for both tables

- [x] In sync, every bet, result set, vote `amounts` goes into eventLeaderboard `investment`

- [x] In sync, when event goes to `WITHDRAWING` status, update `eventLeaderboard winnings and returnRatio`

- [x] In sync,  when event goes to `WITHDRAWING` status, update `globalLeaderboard investments, winnings and returnRatio`

- [x] Migration

- [x] Testings

- [x] Queries

PR #268, opened 2019-07-03
Leaderboard table

BernardHan:
#267 #265 

- [x] Create eventLeaderboard and globalLeaderboard tables and models

- [x] Make db helper methods for both tables

- [x] In sync, every bet, result set, vote `amounts` goes into eventLeaderboard `investment`

- [x] In sync, when event goes to `WITHDRAWING` status, update `eventLeaderboard winnings and returnRatio`

- [x] In sync,  when event goes to `WITHDRAWING` status, update `globalLeaderboard investments, winnings and returnRatio`

- [x] Migration

- [x] Testings

- [x] Queries

### DIFF
--- a/src/db/db-helper.js
+++ b/src/db/db-helper.js
@@ -1,9 +1,12 @@
-const { isNull } = require('lodash');
+const { isNull, each, find } = require('lodash');
 const logger = require('../utils/logger');
 const { isDefined } = require('../utils');
 const { EVENT_STATUS, TX_STATUS } = require('../constants');
 const { db } = require('.');
 const web3 = require('../web3');
+const MultipleResultsEventApi = require('../api/multiple-results-event');
+const EventLeaderboard = require('../models/event-leaderboard');
+const GlobalLeaderboard = require('../models/global-leaderboard');
 
 module.exports = class DBHelper {
   /* Blocks */
@@ -242,7 +245,7 @@ module.exports = class DBHelper {
 
   static async updateEventStatusWithdrawing(currBlockTime) {
     try {
-      await db.Events.update(
+      const updatedEvents = await db.Events.update(
         {
           txStatus: TX_STATUS.SUCCESS,
           $not: { status: EVENT_STATUS.WITHDRAWING },
@@ -250,10 +253,80 @@ module.exports = class DBHelper {
           currentRound: { $gt: 0 },
         },
         { $set: { status: EVENT_STATUS.WITHDRAWING } },
-        { multi: true },
+        {
+          multi: true,
+          returnUpdatedDocs: true,
+        },
       );
+      await DBHelper.updateLeaderboardWithdrawing(updatedEvents);
     } catch (err) {
       logger.error(`UPDATE Event Status Withdrawing error: ${err.message}`);
+      throw err;
+    }
+  }
+
+  static async updateLeaderboardWithdrawing(events) {
+    try {
+      each(events, async (event) => {
+        // get all the participants
+        const bets = await DBHelper.findBet({ eventAddress: event.address });
+        const filtered = [];
+        each(bets, (bet) => {
+          if (!find(filtered, {
+            eventAddress: bet.eventAddress,
+            betterAddress: bet.betterAddress,
+          })) {
+            filtered.push(bet);
+          }
+        });
+        // calculate winnings for all participants
+        const promises = [];
+        const winnings = [];
+        each(filtered, (bet) => {
+          promises.push(new Promise(async (resolve, reject) => {
+            try {
+              const { eventAddress, betterAddress } = bet;
+              const res = await MultipleResultsEventApi.calculateWinnings({
+                eventAddress,
+                address: betterAddress,
+              });
+              winnings.push({
+                eventAddress,
+                betterAddress,
+                amount: res.toString(10),
+              });
+              resolve();
+            } catch (err) {
+              reject(err);
+            }
+          }));
+        });
+        await Promise.all(promises);
+        // update leaderboard
+        each(winnings, async (winning) => {
+          // update event leaderboard winnings
+          const eventLeaderboardEntry = new EventLeaderboard({
+            eventAddress: winning.eventAddress,
+            userAddress: winning.betterAddress,
+            investments: 0,
+            winnings: winning.amount,
+          });
+          await DBHelper.insertEventLeaderboard(eventLeaderboardEntry);
+          // update global leaderboard investments, winnings
+          const userAmountInEvent = await DBHelper.findOneEventLeaderboard({
+            eventAddress: winning.eventAddress,
+            userAddress: winning.betterAddress,
+          });
+          const globalLeaderboard = new GlobalLeaderboard({
+            userAddress: winning.betterAddress,
+            investments: userAmountInEvent.investments,
+            winnings: winning.amount,
+          });
+          await DBHelper.insertGlobalLeaderboard(globalLeaderboard);
+        });
+      });
+    } catch (err) {
+      logger.error(`UPDATE leaderboard when withdrawing status changed error: ${err.message}`);
       throw err;
     }
   }
--- a/src/db/db-helper.js
+++ b/src/db/db-helper.js
@@ -440,4 +440,73 @@ module.exports = class DBHelper {
       throw err;
     }
   }
+
+  /* EventLeaderboard */
+  static async findEventLeaderboard(query, sort) {
+    try {
+      if (sort) return db.EventLeaderboard.cfind(query).sort(sort).exec();
+      return db.EventLeaderboard.find(query);
+    } catch (err) {
+      logger.error(`FIND EventLeaderboard error: ${err.message}`);
+      throw err;
+    }
+  }
+
+  static async findOneEventLeaderboard(query) {
+    try {
+      return db.EventLeaderboard.findOne(query);
+    } catch (err) {
+      logger.error(`FINDONE EventLeaderboard error: ${err.message}`);
+      throw err;
+    }
+  }
+
+  static async countEventLeaderboard(query) {
+    try {
+      return db.EventLeaderboard.count(query);
+    } catch (err) {
+      logger.error(`COUNT EventLeaderboard error: ${err.message}`);
+      throw err;
+    }
+  }
+
+  static async insertEventLeaderboard(entry) {
+    try {
+      const existing = await DBHelper.findOneEventLeaderboard({
+        userAddress: entry.userAddress,
+        eventAddress: entry.eventAddress,
+      });
+      if (isNull(existing)) {
+        await db.EventLeaderboard.insert(entry);
+      } else {
+        await DBHelper.updateEventLeaderboard(existing, entry);
+      }
+    } catch (err) {
+      logger.error(`INSERT EventLeaderboard error: ${err.message}`);
+      throw err;
+    }
+  }
+
+  static async updateEventLeaderboard(existing, entry) {
+    const investments = existing.investments + entry.investments;
+    const winnings = existing.winnings + entry.winnings;
+    try {
+      await db.EventLeaderboard.update(
+        {
+          userAddress: entry.userAddress,
+          eventAddress: entry.eventAddress,
+        },
+        {
+          $set: {
+            investments,
+            winnings,
+            returnRatio: winnings / investments,
+          },
+        },
+      );
+    } catch (err) {
+      logger.error(`UPDATE EventLeaderboard error: ${err.message}`);
+      throw err;
+    }
+  }
 };
--- a/src/db/db-helper.js
+++ b/src/db/db-helper.js
@@ -509,4 +509,67 @@ module.exports = class DBHelper {
       throw err;
     }
   }
+
+  /* GlobalLeaderboard */
+  static async findGlobalLeaderboard(query, sort) {
+    try {
+      if (sort) return db.GlobalLeaderboard.cfind(query).sort(sort).exec();
+      return db.GlobalLeaderboard.find(query);
+    } catch (err) {
+      logger.error(`FIND GlobalLeaderboard error: ${err.message}`);
+      throw err;
+    }
+  }
+
+  static async findOneGlobalLeaderboard(query) {
+    try {
+      return db.GlobalLeaderboard.findOne(query);
+    } catch (err) {
+      logger.error(`FINDONE GlobalLeaderboard error: ${err.message}`);
+      throw err;
+    }
+  }
+
+  static async countGlobalLeaderboard(query) {
+    try {
+      return db.GlobalLeaderboard.count(query);
+    } catch (err) {
+      logger.error(`COUNT GlobalLeaderboard error: ${err.message}`);
+      throw err;
+    }
+  }
+
+  static async insertGlobalLeaderboard(entry) {
+    try {
+      const existing = await DBHelper.findOneGlobalLeaderboard({ userAddress: entry.userAddress });
+      if (isNull(existing)) {
+        await db.GlobalLeaderboard.insert(entry);
+      } else {
+        await DBHelper.updateGlobalLeaderboard(existing, entry);
+      }
+    } catch (err) {
+      logger.error(`INSERT GlobalLeaderboard error: ${err.message}`);
+      throw err;
+    }
+  }
+
+  static async updateGlobalLeaderboard(existing, entry) {
+    const investments = existing.investments + entry.investments;
+    const winnings = existing.winnings + entry.winnings;
+    try {
+      await db.GlobalLeaderboard.update(
+        { userAddress: entry.userAddress },
+        {
+          $set: {
+            investments,
+            winnings,
+            returnRatio: winnings / investments,
+          },
+        },
+      );
+    } catch (err) {
+      logger.error(`UPDATE GlobalLeaderboard error: ${err.message}`);
+      throw err;
+    }
+  }
 };
--- a/src/db/db-helper.js
+++ b/src/db/db-helper.js
@@ -1,6 +1,7 @@
 const { isNull, each, find } = require('lodash');
 const logger = require('../utils/logger');
 const { isDefined } = require('../utils');
+const { sumBN } = require('../utils/web3-utils');
 const { EVENT_STATUS, TX_STATUS } = require('../constants');
 const { db } = require('.');
 const web3 = require('../web3');
@@ -553,7 +554,15 @@ module.exports = class DBHelper {
       if (isNull(existing)) {
         await db.EventLeaderboard.insert(entry);
       } else {
-        await DBHelper.updateEventLeaderboard(existing, entry);
+        const investments = sumBN(existing.investments, entry.investments).toString(10);
+        const winnings = sumBN(existing.winnings, entry.winnings).toString(10);
+        const newEntry = new EventLeaderboard({
+          userAddress: entry.userAddress,
+          eventAddress: entry.eventAddress,
+          investments,
+          winnings,
+        }); // make a new instance, ratio will be calculated in the model
+        await DBHelper.updateEventLeaderboard(newEntry);
       }
     } catch (err) {
       logger.error(`INSERT EventLeaderboard error: ${err.message}`);
@@ -561,24 +570,14 @@ module.exports = class DBHelper {
     }
   }
 
-  static async updateEventLeaderboard(existing, entry) {
-    const existingInvest = web3.utils.toBN(existing.investments);
-    const investments = existingInvest.add(web3.utils.toBN(entry.investments)).toString(10);
-    const existingWin = web3.utils.toBN(existing.winnings);
-    const winnings = existingWin.add(web3.utils.toBN(entry.winnings)).toString(10);
+  static async updateEventLeaderboard(entry) {
     try {
       await db.EventLeaderboard.update(
         {
           userAddress: entry.userAddress,
           eventAddress: entry.eventAddress,
         },
-        {
-          $set: {
-            investments,
-            winnings,
-            returnRatio: winnings / investments,
-          },
-        },
+        { $set: entry },
       );
     } catch (err) {
       logger.error(`UPDATE EventLeaderboard error: ${err.message}`);
@@ -621,7 +620,14 @@ module.exports = class DBHelper {
       if (isNull(existing)) {
         await db.GlobalLeaderboard.insert(entry);
       } else {
-        await DBHelper.updateGlobalLeaderboard(existing, entry);
+        const investments = sumBN(existing.investments, entry.investments).toString(10);
+        const winnings = sumBN(existing.winnings, entry.winnings).toString(10);
+        const newEntry = new GlobalLeaderboard({
+          userAddress: entry.userAddress,
+          investments,
+          winnings,
+        }); // make a new instance, ratio will be calculated in the model
+        await DBHelper.updateGlobalLeaderboard(newEntry);
       }
     } catch (err) {
       logger.error(`INSERT GlobalLeaderboard error: ${err.message}`);
@@ -629,21 +635,11 @@ module.exports = class DBHelper {
     }
   }
 
-  static async updateGlobalLeaderboard(existing, entry) {
-    const existingInvest = web3.utils.toBN(existing.investments);
-    const investments = existingInvest.add(web3.utils.toBN(entry.investments)).toString(10);
-    const existingWin = web3.utils.toBN(existing.winnings);
-    const winnings = existingWin.add(web3.utils.toBN(entry.winnings)).toString(10);
+  static async updateGlobalLeaderboard(entry) {
     try {
       await db.GlobalLeaderboard.update(
         { userAddress: entry.userAddress },
-        {
-          $set: {
-            investments,
-            winnings,
-            returnRatio: winnings / investments,
-          },
-        },
+        { $set: entry },
       );
     } catch (err) {
       logger.error(`UPDATE GlobalLeaderboard error: ${err.message}`);
--- a/src/db/db-helper.js
+++ b/src/db/db-helper.js
@@ -3,6 +3,7 @@ const logger = require('../utils/logger');
 const { isDefined } = require('../utils');
 const { EVENT_STATUS, TX_STATUS } = require('../constants');
 const { db } = require('.');
+const web3 = require('../web3');
 
 module.exports = class DBHelper {
   /* Blocks */
@@ -488,8 +489,10 @@ module.exports = class DBHelper {
   }
 
   static async updateEventLeaderboard(existing, entry) {
-    const investments = existing.investments + entry.investments;
-    const winnings = existing.winnings + entry.winnings;
+    const existingInvest = web3.utils.toBN(existing.investments);
+    const investments = existingInvest.add(web3.utils.toBN(entry.investments)).toString(10);
+    const existingWin = web3.utils.toBN(existing.winnings);
+    const winnings = existingWin.add(web3.utils.toBN(entry.winnings)).toString(10);
     try {
       await db.EventLeaderboard.update(
         {
@@ -554,8 +557,10 @@ module.exports = class DBHelper {
   }
 
   static async updateGlobalLeaderboard(existing, entry) {
-    const investments = existing.investments + entry.investments;
-    const winnings = existing.winnings + entry.winnings;
+    const existingInvest = web3.utils.toBN(existing.investments);
+    const investments = existingInvest.add(web3.utils.toBN(entry.investments)).toString(10);
+    const existingWin = web3.utils.toBN(existing.winnings);
+    const winnings = existingWin.add(web3.utils.toBN(entry.winnings)).toString(10);
     try {
       await db.GlobalLeaderboard.update(
         { userAddress: entry.userAddress },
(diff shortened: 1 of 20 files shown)


src/models/event-leaderboard.js
```diff
@@ -0,0 +1,25 @@
...
+    if (!isString(params.investments)) throw Error('investments must be a Number');
+    if (!isFinite(params.winnings)) throw Error('winnings must be a Number');
+  }
+
+  format(params) {
+    this.eventAddress = toLowerCase(params.eventAddress);
+    this.userAddress = toLowerCase(params.userAddress);
+    this.investments = params.investments;
+    this.winnings = params.winnings;
+    this.returnRatio = this.winnings / this.investments;
```
dwalintukan: cant do this with two satoshi strings so you need to convert to BN. but there is a caveat, you cant just divide it as BNs since it does not handle floating point precision.

so Web3 seems like they are considering BigNumber.js as a temp for handling this, but the reason they took it out was for performance issues (https://github.com/ethereum/web3.js/issues/2171).

I guess for our case, we can use BigNumber for this specific case to handle the precision and percentages.

src/models/global-leaderboard.js
```diff
@@ -0,0 +1,23 @@
...
+
+module.exports = class GlobalLeaderboard {
+  constructor(params) {
+    this.validate(params);
+    this.format(params);
+  }
+
+  validate(params) {
+    if (!isString(params.userAddress)) throw Error('userAddress must be a String');
+    if (!isString(params.investments)) throw Error('investments must be a Number');
```
dwalintukan: must be a String

src/models/global-leaderboard.js
```diff
@@ -0,0 +1,23 @@
...
+    if (!isFinite(params.winnings)) throw Error('winnings must be a Number');
```
dwalintukan: use a Satoshi string here

src/models/event-leaderboard.js
```diff
@@ -0,0 +1,25 @@
+/* eslint no-underscore-dangle: 0 */
+const { isFinite, isString } = require('lodash');
+const { toLowerCase } = require('../utils');
+
+module.exports = class GlobalLeaderboard {
```
dwalintukan: EventLeaderboard

src/models/event-leaderboard.js
```diff
@@ -0,0 +1,25 @@
+/* eslint no-underscore-dangle: 0 */
```
dwalintukan: dont have underscore dangles here

src/models/global-leaderboard.js
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  format(params) {
+    this.userAddress = toLowerCase(params.userAddress);
+    this.investments = params.investments;
+    this.winnings = params.winnings;
+    this.returnRatio = this.winnings / this.investments;
```
dwalintukan: same comment as EventLeaderboard. lets try to use BigNumber to get the decimal ratio here

src/models/global-leaderboard.js
```diff
@@ -0,0 +1,23 @@
+/* eslint no-underscore-dangle: 0 */
```
dwalintukan: no underscore dangle

src/db/db-helper.js
```diff
@@ -241,22 +245,92 @@ module.exports = class DBHelper {
...
+        },
       );
+      await DBHelper.updateLeaderboardWithdrawing(updatedEvents);
     } catch (err) {
       logger.error(`UPDATE Event Status Withdrawing error: ${err.message}`);
       throw err;
     }
   }
 
+  static async updateLeaderboardWithdrawing(events) {
```
dwalintukan: move this logic to a separate sync file like how each sync step has its own. keep db helper clean with just db operations.

src/db/db-helper.js
```diff
@@ -241,22 +245,92 @@ module.exports = class DBHelper {
 
   static async updateEventStatusWithdrawing(currBlockTime) {
     try {
-      await db.Events.update(
+      const updatedEvents = await db.Events.update(
```
dwalintukan: return the updatedDocs and then pass it to the sync logic for updateLeaderboardWithdrawing

src/db/db-helper.js
```diff
@@ -241,22 +245,92 @@ module.exports = class DBHelper {
...
+            betterAddress: bet.betterAddress,
+          })) {
+            filtered.push(bet);
+          }
+        });
+        // calculate winnings for all participants
+        const promises = [];
+        const winnings = [];
+        each(filtered, (bet) => {
+          promises.push(new Promise(async (resolve, reject) => {
```
dwalintukan: this needs to be batched up. our RPC server maxes out at ~40 requests per second right now. if you were parsing 4 events x 10 winners per event, then you'd max out.

I am using `pLimit` (https://github.com/sindresorhus/p-limit) to handle this case. take a look how its currently implemented right now. use the same `limit` var in `sync/index`. this will work out when you move this logic into its own sync file.

https://github.com/bodhiproject/bodhi-server/blob/dev-062419/src/sync/index.js#L31
https://github.com/bodhiproject/bodhi-server/blob/dev-062419/src/sync/index.js#L183
https://github.com/bodhiproject/bodhi-server/blob/dev-062419/src/sync/bet-placed.js#L24

src/db/db-helper.js
```diff
@@ -241,22 +245,92 @@ module.exports = class DBHelper {
...
+              });
+              resolve();
+            } catch (err) {
+              reject(err);
+            }
+          }));
+        });
+        await Promise.all(promises);
+        // update leaderboard
+        each(winnings, async (winning) => {
```
dwalintukan: move all the DB updating inside the promises above so one promise takes care of everything for one user, API call + insertion. why do you need to loop twice?

src/db/db-helper.js
```diff
@@ -241,22 +245,92 @@ module.exports = class DBHelper {
...
+          // update event leaderboard winnings
+          const eventLeaderboardEntry = new EventLeaderboard({
+            eventAddress: winning.eventAddress,
+            userAddress: winning.betterAddress,
+            investments: 0,
```
dwalintukan: This is always 0. You should handle this when you fetch all the bets for this event on L274 by accumulating all the bets for each individual user and just stick it here.

src/db/db-helper.js
```diff
@@ -241,22 +245,92 @@ module.exports = class DBHelper {
...
     }
   }
 
+  static async updateLeaderboardWithdrawing(events) {
+    try {
+      each(events, async (event) => {
+        // get all the participants
+        const bets = await DBHelper.findBet({ eventAddress: event.address });
+        const filtered = [];
+        each(bets, (bet) => {
```
dwalintukan: 1. use lodash's reduce to accumulate all the bets in order to get the `investment`
2. keep track of the winners only. in the loop below this, you are calling `calculateWinnings` on the loser's bets as well so it's making unnecessary calls.

src/db/db-helper.js
```diff
@@ -241,22 +245,92 @@ module.exports = class DBHelper {
...
+          // update event leaderboard winnings
+          const eventLeaderboardEntry = new EventLeaderboard({
+            eventAddress: winning.eventAddress,
+            userAddress: winning.betterAddress,
+            investments: 0,
+            winnings: winning.amount,
+          });
+          await DBHelper.insertEventLeaderboard(eventLeaderboardEntry);
+          // update global leaderboard investments, winnings
+          const userAmountInEvent = await DBHelper.findOneEventLeaderboard({
```
dwalintukan: unnecessary lookup. why fetch the same obj after inserting it?

src/db/db-helper.js
```diff
@@ -440,4 +514,140 @@ module.exports = class DBHelper {
...
+        await DBHelper.updateGlobalLeaderboard(existing, entry);
+      }
+    } catch (err) {
+      logger.error(`INSERT GlobalLeaderboard error: ${err.message}`);
+      throw err;
+    }
+  }
+
+  static async updateGlobalLeaderboard(existing, entry) {
+    const existingInvest = web3.utils.toBN(existing.investments);
```
dwalintukan: move these 4 calculations to the `else` block in `insertGlobalLeaderboard` and set it on the existing obj then just pass the updated obj to `updateGlobalLeaderboard`

src/db/db-helper.js
```diff
@@ -440,4 +514,140 @@ module.exports = class DBHelper {
...
+        await DBHelper.updateEventLeaderboard(existing, entry);
+      }
+    } catch (err) {
+      logger.error(`INSERT EventLeaderboard error: ${err.message}`);
+      throw err;
+    }
+  }
+
+  static async updateEventLeaderboard(existing, entry) {
+    const existingInvest = web3.utils.toBN(existing.investments);
```
dwalintukan: move these 4 calculations to the `else` block in `insertEventLeaderboard` and pass the updated entry to this function only

src/db/db-helper.js
```diff
@@ -440,4 +514,140 @@ module.exports = class DBHelper {
...
+    const investments = existingInvest.add(web3.utils.toBN(entry.investments)).toString(10);
```
dwalintukan: use `sumBN` in `utils/web3-utils`

src/db/db-helper.js
```diff
@@ -440,4 +514,140 @@ module.exports = class DBHelper {
...
+    const existingWin = web3.utils.toBN(existing.winnings);
+    const winnings = existingWin.add(web3.utils.toBN(entry.winnings)).toString(10);
+    try {
+      await db.EventLeaderboard.update(
+        {
+          userAddress: entry.userAddress,
+          eventAddress: entry.eventAddress,
+        },
+        {
+          $set: {
```
dwalintukan: `$set: entry` after changing it to pass the entire updated object. this makes this func reusable other than just these three fields (if ever needed).

src/db/db-helper.js
```diff
@@ -440,4 +514,140 @@ module.exports = class DBHelper {
...
+  static async updateGlobalLeaderboard(existing, entry) {
+    const existingInvest = web3.utils.toBN(existing.investments);
+    const investments = existingInvest.add(web3.utils.toBN(entry.investments)).toString(10);
+    const existingWin = web3.utils.toBN(existing.winnings);
+    const winnings = existingWin.add(web3.utils.toBN(entry.winnings)).toString(10);
+    try {
+      await db.GlobalLeaderboard.update(
+        { userAddress: entry.userAddress },
+        {
+          $set: {
```
dwalintukan: `$set: entry`

src/db/index.js
```diff
@@ -37,14 +41,18 @@ const initDB = async () => {
...
+      db.EventLeaderboard.loadDatabase(),
     ]);
 
     await db.Blocks.ensureIndex({ fieldName: 'blockNum', unique: true });
     await db.Events.ensureIndex({ fieldName: 'txid', unique: true });
     await db.Bets.ensureIndex({ fieldName: 'txid', unique: true });
     await db.ResultSets.ensureIndex({ fieldName: 'txid', unique: true });
     await db.Withdraws.ensureIndex({ fieldName: 'txid', unique: true });
-    await db.TransactionReceipts.ensureIndex({ fieldName: 'transactionHash', unique: true });
+    // await db.TransactionReceipts.ensureIndex({ fieldName: 'transactionHash', unique: true });
```
dwalintukan: Undo this. I synced from scratch today and did not run into this problem. Chalking it up to bad data.

src/db/db-helper.js
```diff
@@ -440,4 +514,140 @@ module.exports = class DBHelper {
...
+            investments,
+            winnings,
+            returnRatio: winnings / investments,
```
dwalintukan: this shouldnt work, see my comments in the models

src/db/db-helper.js
```diff
@@ -440,4 +514,140 @@ module.exports = class DBHelper {
...
+      await db.EventLeaderboard.update(
+        {
+          userAddress: entry.userAddress,
+          eventAddress: entry.eventAddress,
+        },
+        {
+          $set: {
+            investments,
+            winnings,
+            returnRatio: winnings / investments,
```
dwalintukan: this shouldnt work, see comments in models

src/sync/bet-placed.js
```diff
@@ -72,6 +82,14 @@ const pendingBetPlaced = async ({ syncPromises, limit }) => {
             if (foundLog) {
               const bet = parseBet({ log: foundLog });
               await DBHelper.insertBet(bet);
+              // Update event leaderboard investments
```
dwalintukan: 👾  above

src/db/db-helper.js
BernardHan: 0 here is just to avoid duplicate addings. Event leaderboard entry should already have the investment for this user.

src/sync/bet-placed.js
BernardHan: what? lol

src/models/global-leaderboard.js
```diff
@@ -0,0 +1,24 @@
...
+    if (!isString(params.investments)) throw Error('investments must be a String');
+    if (!isString(params.winnings)) throw Error('winnings must be a String');
+  }
+
+  format(params) {
+    this.userAddress = toLowerCase(params.userAddress);
+    this.investments = params.investments;
+    this.winnings = params.winnings;
+    const ratio = new BigNumber(this.winnings).dividedBy(new BigNumber(this.investments));
+    this.returnRatio = ratio.toString();
```
dwalintukan: do we need to convert to a string? wouldnt a number suffice here since were 0-100?

src/models/event-leaderboard.js
```diff
@@ -0,0 +1,26 @@
...
+    if (!isString(params.winnings)) throw Error('winnings must be a String');
+  }
+
+  format(params) {
+    this.eventAddress = toLowerCase(params.eventAddress);
+    this.userAddress = toLowerCase(params.userAddress);
+    this.investments = params.investments;
+    this.winnings = params.winnings;
+    const ratio = new BigNumber(this.winnings).dividedBy(new BigNumber(this.investments));
+    this.returnRatio = ratio.toString();
```
dwalintukan: same question. number instead of string?

src/db/index.js
dwalintukan: needs rebase, Henry merged in his PR regarding this line

src/sync/bet-placed.js
dwalintukan: its a space invader (https://en.wikipedia.org/wiki/Space_Invaders), ya prob before your time lol
means either add whitespace above, or remove it if there was an extra line

src/sync/event-status-withdrawing-changed.js
```diff
@@ -0,0 +1,112 @@
...
+    }
+    return acc;
+  }, {});
+
+  return accumulated;
+};
+
+const updateLeaderboardWithdrawing = async (syncPromises, events, limit) => {
+  each(events, async (event) => {
+    syncPromises.push(limit(async () => {
```
dwalintukan: need to pass `event` to the `limit(...func, event)`. see https://github.com/bodhiproject/bodhi-server/pull/270/files#diff-ba9f3a5bc0319fb93654cb9d99155063R24

src/sync/event-status-withdrawing-changed.js
```diff
@@ -0,0 +1,112 @@
...
+        });
+        // calculate winnings for all participants
+        const promises = [];
+        each(filtered, (tx) => {
+          promises.push(new Promise(async (resolve, reject) => {
+            try {
+              const { eventAddress, betterAddress, resultIndex, centralizedOracleAddress } = tx;
+              const userAddress = betterAddress || centralizedOracleAddress;
+              // calculate winning for this user in this event
+              let winning = '0';
```
dwalintukan: nit: `winnings` so you can shortform it

src/sync/event-status-withdrawing-changed.js
```diff
@@ -0,0 +1,112 @@
...
+              if (resultIndex === event.currentResultIndex) {
+                const res = await MultipleResultsEventApi.calculateWinnings({
```
dwalintukan: we would still have the parallel limit issue here since the `limit()` is only for the outer promise. let me dig into this issue and lets leave this for now.

src/sync/index.js
```diff
@@ -205,7 +206,10 @@ const startSync = async () => {
     await DBHelper.updateEventStatusOracleResultSetting(blockTime);
     await DBHelper.updateEventStatusOpenResultSetting(blockTime);
     await DBHelper.updateEventStatusArbitration(blockTime);
-    await DBHelper.updateEventStatusWithdrawing(blockTime);
+    // await DBHelper.updateEventStatusWithdrawing(blockTime);
```
dwalintukan: undo this and keep it separate. set a var newWithdrawEvents then pass it to `updateLeaderboards({ newWithdrawEvents, syncPromises, limit })`

src/models/global-leaderboard.js
BernardHan: The result will be a double type, think it better be string?

src/sync/event-status-withdrawing-changed.js
BernardHan: ok will do.

````Support for passing arguments on to the fn is provided in order to be able to avoid creating unnecessary closures. You probably don't need this optimization unless you're pushing a lot of functions.```

But I don't quite understand what will happen if no such optimization.

File: src/sync/event-status-withdrawing-changed.js (hunk 28 above)
Comment (BernardHan) [reply src/sync/event-status-withdrawing-changed.js]: can we move the limit to wrap this part so no inner promises?
We query the bets and resultsets first, then start to limit.

File: src/models/global-leaderboard.js (hunk 24 above)
Comment (dwalintukan) [reply src/models/global-leaderboard.js]: We should keep the types as native as possible. Percentages are number types. The only reason we use Strings for other places is because it would overflow due to being too large.

File: src/sync/event-status-withdrawing-changed.js (hunk 26 above)
Comment (dwalintukan) [reply src/sync/event-status-withdrawing-changed.js]: It seems to be primarily for reference issues. In this situation, the promises are holding a ref to the `event` outside of promise itself. So in regards to garbage collection, the promise it forcing it to hold an a reference outside of the promise.

If you passed the event to the limit func, it wouldn't need that ref outside of it since its being passed in.

Comment (dwalintukan) [reply src/sync/event-status-withdrawing-changed.js]: Also, I verified that you dont need to pass funcs to the limit func since any funcs called within the limit func will inherently be limited as well.

File: src/db/index.js (hunk 20 above)
Comment (dwalintukan) [reply src/db/index.js]: Why still commented? Henry just removed the `unique` prop of it.

File: src/graphql/queries/leaderboard.js [hunk 30]
```diff
@@ -0,0 +1,53 @@
+const { isArray, each } = require('lodash');
+const { lowercaseFilters, runPaginatedQuery } = require('./utils');
+
+const buildFilters = ({
+  userAddress,
+  eventAddress,
+} = {}) => {
+  if (!eventAddress && !userAddress) {
+    throw Error('eventAddress or userAddress missing in filters');
```
dwalintukan: or -> and?

src/graphql/queries/leaderboard.js
```diff
@@ -0,0 +1,53 @@
+const { isArray, each } = require('lodash');
+const { lowercaseFilters, runPaginatedQuery } = require('./utils');
+
+const buildFilters = ({
+  userAddress,
+  eventAddress,
+} = {}) => {
+  if (!eventAddress && !userAddress) {
```
dwalintukan: this should be `||`? since GlobalLeaderboard can be queried by userAddress only?

src/graphql/queries/leaderboard.js
```diff
@@ -0,0 +1,53 @@
...
+  filters.push(filter);
+  return filters;
+};
+
+const eventLeaderboard = async (
+  root,
+  { filter, orderBy, limit, skip },
+  { db: { EventLeaderboard } },
+) => {
+  const query = filter ? { $or: buildFilters(lowercaseFilters(filter)) } : {};
```
dwalintukan: you need to add `userAddress` or use one of the fields that are already in there like `betterAddress` in the `lowercaseFilters` func

src/graphql/schema.js
```diff
@@ -239,11 +239,30 @@ type PaginatedBiggestWinner {
   items: [BiggestWinner]!
 }
 
+type Leaderboard {
+  eventAddress: String
+  userAddress: String!
+  investments: String!
+  winnings: String!
+  returnRatio: String!
```
dwalintukan: change to Int

src/graphql/schema.js
```diff
@@ -239,11 +239,30 @@ type PaginatedBiggestWinner {
   items: [BiggestWinner]!
 }
 
+type Leaderboard {
```
dwalintukan: `LeaderboardUser`? Leaderboard is the entity of the entire board itself so need to differentiate individual items.

src/graphql/schema.js
```diff
@@ -239,11 +239,30 @@ type PaginatedBiggestWinner {
...
+  eventAddress: String
+  userAddress: String!
+  investments: String!
+  winnings: String!
+  returnRatio: String!
+}
+
+type PaginatedLeaderboard {
```
dwalintukan: PaginatedLeaderboardUser

src/sync/event-status-withdrawing-changed.js
dwalintukan: Lets use https://caolan.github.io/async/v3/docs.html#eachOfSeries to handle the `calculateWinnings` and keep the outer limit promise. Would prefer not to limit the logic on the inside of an already parallel promise. Not too familiar with the logic of limit to start placing it in async parallel execution. Not too sure what would happen.

This will be slower since its running each calcWinnings individually but this ensures the concurrency limit is not broken.

src/graphql/queries/leaderboard.js
BernardHan: What i have right now means `if no eventAddress and no userAddress then error`, which is what we need right?

No matter which one, we need to pass at least one of them.

src/graphql/schema.js
BernardHan: Think percentage should be a float number right? I can change it to Float

src/sync/update-leaderboard.js
```diff
@@ -51,44 +88,21 @@ const updateLeaderboardWithdrawing = async (syncPromises, events, limit) => {
...
-                investments: investments[userAddress],
-                winnings,
-              });
-              await DBHelper.insertGlobalLeaderboard(globalLeaderboard);
-              resolve();
-            } catch (err) {
-              reject(err);
-            }
-          }));
+        async.forEachOf(filtered, (value, key, callback) => {
```
BernardHan: forgot to change to use `eachOf`. Will change and test later.

src/graphql/queries/leaderboard.js
dwalintukan: gotcha, ya thats what we need.

src/graphql/schema.js
dwalintukan: sorry ya Float, not Int. we want to retain 2 decimal places.

src/graphql/schema.js
```diff
@@ -418,19 +418,19 @@ type Query {
     skip: Int
   ): PaginatedBiggestWinner!
 
-  eventLeaderboard(
-    filter: LeaderboardFilter
+  eventLeaderboardEntries(
```
dwalintukan: This should still be `eventLeaderboard`.  Sorry wasn't clear. The `eventLeaderboard` query returns a list of `LeaderboardEntry`s. Like a `users` query will return a list of `User`s. 

The relation is one-to-many. One `eventLeaderboard` query will return many `LeaderboardEntry`s.

src/graphql/schema.js
```diff
@@ -418,19 +418,19 @@ type Query {
...
+    filter: LeaderboardEntryFilter
     orderBy: [Order!]
     limit: Int
     skip: Int
-  ): PaginatedLeaderboard!
+  ): PaginatedLeaderboardEntry!
 
-  globalLeaderboard(
-    filter: LeaderboardFilter
+  globalLeaderboardEntries(
```
dwalintukan: same thing here

src/sync/update-leaderboard.js
dwalintukan: please change before this gets merged, `eachOf` still is doing parallel exec.

src/graphql/schema.js
BernardHan: Think `eventLeaderboardEntries` works better since we are querying for a set of paginated entries not the whole leaderboard.

In Event store, we only query for one specific entry just to get the winnings.

dwalintukan: Fair enough. Actually it does conform to the other query names.

src/graphql/queries/leaderboard.js
```diff
@@ -23,13 +24,20 @@ const eventLeaderboardEntries = async (
...
     db: EventLeaderboard,
     filter: query,
     orderBy,
     limit,
     skip,
   });
+
+  if (orderBy && orderBy.length > 0) {
+    const { field } = orderBy[0];
+    res.items.sort((a, b) => new BigNumber(b[field]).comparedTo(new BigNumber(a[field]))); // all descending
```
dwalintukan: what if trying to sort by address?

dwalintukan: nit: suggest to maybe use `BN.js` in this case. not sure if the performance impacts with `BigNumber` would come into play here.

BernardHan: `field` here is either `investments` or `winnings`, which are all converted from `BigNumber` instances, think it's better to keep it consistent to use `BigNumber`?

dwalintukan: 1. Need to handle what happens if  someone passes in `{ field: 'eventAddress', direction: 'ASC' }` or `betterAddress`. It will crash here since you're trying to convert an address to a BigNumber.
2. Was just worried about the performance of it. If you feel the performance is good compared to BN, then let's just keep it BigNumber. But would like to still use BN whenever possible to keep it standardized throughout the code.

test/db/index.js
```diff
@@ -1771,6 +1771,384 @@ describe('db', () => {
...
+        investments: '2000000',
+        winnings: '0',
+        returnRatio: 0,
+      },
+    ];
+    let eventLeaderboardEntry;
+    let eventLeaderboardEntryCount;
+
+    describe('findEventLeaderboard', () => {
+      it('find empty withdraw db', async () => {
```
dwalintukan: double check when you are copy/pasting code

test/db/index.js
```diff
@@ -1771,6 +1771,384 @@ describe('db', () => {
...
+        });
+        should.exist(eventLeaderboardEntry);
+        expect(eventLeaderboardEntry).excluding(['_id']).to.deep.equal(expectedEntry);
+      });
+
+      it('should update winnings if insert existed entry', async () => {
+        const newEntry = {
+          eventAddress: '0x12345ea6e4e1f5375f7596b73b3b597e6507201a',
+          userAddress: '0x123456764c0bd3355b2d54e4fa2203e8343b6d6a',
+          investments: '0',
```
dwalintukan: shouldn't the `investments` be the same?

test/db/index.js
```diff
@@ -1771,6 +1771,384 @@ describe('db', () => {
...
+        eventLeaderboardEntryCount.should.equal(3);
+        eventLeaderboardEntry = await DBHelper.findOneEventLeaderboard({
+          userAddress: mockEventLeaderboard[2].userAddress,
+          eventAddress: mockEventLeaderboard[2].eventAddress,
+        });
+        should.exist(eventLeaderboardEntry);
+        expect(eventLeaderboardEntry).excluding(['_id']).to.deep.equal(newEntry);
+      });
+
+      it('should not update withdraw when eventAddress is not existed', async () => {
```
dwalintukan: check wording

test/db/index.js
```diff
@@ -1771,6 +1771,384 @@ describe('db', () => {
...
+        eventLeaderboardEntryCount = await db.EventLeaderboard.count({});
+        eventLeaderboardEntryCount.should.equal(3);
+        eventLeaderboardEntry = await DBHelper.findOneEventLeaderboard({
+          userAddress: newEntry.userAddress,
+          eventAddress: newEntry.eventAddress,
+        });
+        should.not.exist(eventLeaderboardEntry);
+      });
+
+      it('should not update withdraw when userAddress is not existed', async () => {
```
dwalintukan: wording

test/db/index.js
```diff
@@ -1771,6 +1771,384 @@ describe('db', () => {
...
+        expect(globalLeaderboardEntry).excluding(['_id']).to.deep.not.equal(newEntry);
+        await DBHelper.updateGlobalLeaderboard(newEntry);
+        globalLeaderboardEntryCount = await db.GlobalLeaderboard.count({});
+        globalLeaderboardEntryCount.should.equal(3);
+        globalLeaderboardEntry = await DBHelper.findOneGlobalLeaderboard({ userAddress: mockGlobalLeaderboard[2].userAddress });
+        should.exist(globalLeaderboardEntry);
+        expect(globalLeaderboardEntry).excluding(['_id']).to.deep.equal(newEntry);
+      });
+
+      it('should not update withdraw when userAddress is not existed', async () => {
```
dwalintukan: wording

src/graphql/queries/leaderboard.js
BernardHan: ratio is float type, think it's better keep it consistent to use Bignumber

I will skip the sorting if it's address

dwalintukan: Ah right, ya if dealing w/ floats, then BigNumber all the way.

PR #273, opened 2019-07-09
Migration

henryzbw:


src/db/migrations/migration2.js
```diff
@@ -0,0 +1,116 @@
...
+          check => check(null, i < txs.length), // trigger iter
+          (next) => {
+            const tx = txs[i];
+            i++;
+            try {
+              const userAddress = tx.betterAddress || tx.centralizedOracleAddress;
+              const eventLeaderboardEntry = new EventLeaderboard({
+                eventAddress,
+                userAddress,
+                investments: tx.amount, // event leaderboard entry already has user's investments
```
BernardHan: remove the copied comment, it's misleading in this migration

src/db/migrations/migration2.js
```diff
@@ -0,0 +1,116 @@
...
+        txs.forEach((tx, index) => {
+          if(tx.resultIndex !== event.currentResultIndex) return;
+          if(tx.betterAddress) addresses.add(tx.betterAddress);
+          if(tx.centralizedOracleAddress) addresses.add(tx.centralizedOracleAddress);
+        })
+        addresses.forEach((address, index) => {
+          promises.push(limit(async () => {
+            try {
+              // calculate winning for this user in this event
+              let winnings = '0';
```
BernardHan: assign winnings after await with `const`, it never gets reassigned

BernardHan:
@dwalintukan may want the `plimit` to wrap the very outer logic, but migration is one time execution, if this does not cause problem, I'd say it looks good to me.

src/db/migrations/migration2.js
```diff
@@ -0,0 +1,116 @@
...
+      }
+      if (event.status === EVENT_STATUS.WITHDRAWING) {
+        const addresses = new Set();
+        txs.forEach((tx, index) => {
+          if(tx.resultIndex !== event.currentResultIndex) return;
+          if(tx.betterAddress) addresses.add(tx.betterAddress);
+          if(tx.centralizedOracleAddress) addresses.add(tx.centralizedOracleAddress);
+        })
+        addresses.forEach((address, index) => {
+          promises.push(limit(async () => {
```
dwalintukan: move this `limit` call to the very outer loop. so 1 promise will take care of 1 event.

this outer loop `events.forEach(async (event, key) => {` will keep running even though the `limit` here will pause operation so it "should" queue everything up anyways once it hits the 15 concurrency limit. prefer to apply the `limit` on the entire event parsing and have each `calculateWinnings` call sequentially. so each `limit` func will correspond to all the inserts + calcWinnings of one event, and calcWinnings will only be called once at a time per event. 

1. move `limit` to very outer loop
2. use `eachOfSeries` in async lib to handle calculateWinnings

src/db/migrations/migration2.js
```diff
@@ -0,0 +1,116 @@
...
+        eventAddress: event.address,
+        txStatus: TX_STATUS.SUCCESS,
+      });
+      const txs = resultSet.concat(bets);
+
+      try {
+        let i = 0;
+        await async.whilst(
+          check => check(null, i < txs.length), // trigger iter
+          (next) => {
```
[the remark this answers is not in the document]
henryzbw: tried declare as async there, will throw error as next is not a function

src/db/migrations/migration2.js
```diff
@@ -0,0 +1,120 @@
...
+          eventAddress: event.address,
+          txStatus: TX_STATUS.SUCCESS,
+        });
+        const txs = resultSet.concat(bets);
+
+        try {
+          let i = 0;
+          await async.whilst(
+            check => check(null, i < txs.length), // trigger iter
+            (next) => {
```
dwalintukan: Rename next, this is shadowing the same var as the `async function migration2(next) {`.

src/db/migrations/migration2.js
```diff
@@ -0,0 +1,120 @@
+const pLimit = require('p-limit');
+const async = require('async');
```
dwalintukan: nit: destructure only funcs that you need

src/db/migrations/migration2.js
dwalintukan: I just tested, you can use `async () => {`. See here for how to implement: https://caolan.github.io/async/v3/global.html

```
await async.whilst(
            check => check(null, i < txs.length), // trigger iter
            async () => {
              const tx = txs[i];
              i++;
              const userAddress = tx.betterAddress || tx.centralizedOracleAddress;
              const eventLeaderboardEntry = new EventLeaderboard({
                eventAddress,
                userAddress,
                investments: tx.amount,
                winnings: '0',
              });
              await DBHelper.insertEventLeaderboard(eventLeaderboardEntry);
              if (event.status === EVENT_STATUS.WITHDRAWING) {
                const globalLeaderboardEntry = new GlobalLeaderboard({
                  userAddress,
                  investments: tx.amount,
                  winnings: '0',
                });
                await DBHelper.insertGlobalLeaderboard(globalLeaderboardEntry);
              }
            },
          );
```

PR #283, opened 2019-07-12
Better err handling

dwalintukan:
# Pull Request Template

## Checklist

## Notes

Will handle the regression tests in a later PR.

### DIFF
--- a/docker/mainnet/docker-compose.yml
+++ b/docker/mainnet/docker-compose.yml
@@ -4,7 +4,7 @@ services:
   bodhi-server-mainnet:
     hostname: bodhi-server-mainnet
     container_name: bodhi-server-mainnet
-    restart: unless-stopped
+    restart: always
     stop_signal: SIGINT
     build:
       context: ../../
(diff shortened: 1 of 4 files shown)


docker/mainnet/docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
   bodhi-server-mainnet:
     hostname: bodhi-server-mainnet
     container_name: bodhi-server-mainnet
-    restart: unless-stopped
+    restart: always
```
BernardHan: Remind me why we had `unless-stopped` at the first place?

src/index.js
```diff
@@ -15,10 +20,9 @@ const start = async () => {
     initSync();
     require('./web3');
     initApi();
-    startSync(true);
+    await startSync(true);
```
BernardHan: is `await` here a must?

docker/mainnet/docker-compose.yml
dwalintukan: ya it wasnt the proper restart policy to begin with. generally we want `always` in all cases to have max uptime.

src/index.js
dwalintukan: i was playing around with getting it to handle unhandled promise exceptions, but found a diff solution for it. but either way its proper to set await on it. if it was returning then we wouldn't have to add the await.

Issue #278, opened 2019-07-12
Remove writing to start_block.dat and just re-parse highest block - BLOCK_BATCH_COUNT


dwalintukan:
This will eliminate the need to write/read to the file. It's essentially the same logic.

PR #285, opened 2019-07-12
278 remove start block file

dwalintukan:
# Pull Request Template

## Checklist

## Notes

#278 
Tests will be included in the full sync tests.

### DIFF
--- a/src/sync/index.js
+++ b/src/sync/index.js
@@ -1,5 +1,6 @@
 const pLimit = require('p-limit');
 const moment = require('moment');
+const { isNumber } = require('lodash');
 const { eventFactoryMeta, isMainnet } = require('../config');
 const { EVENT_MESSAGE } = require('../constants');
 const web3 = require('../web3');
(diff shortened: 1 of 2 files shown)


src/sync/index.js
```diff
@@ -115,14 +66,12 @@ const initSync = () => {
...
-  if (start) return start;
-
   const blocks = await DBHelper.findLatestBlock();
   if (blocks.length > 0) {
-    // Blocks found in DB, use the last synced block as start
-    start = blocks[0].blockNum + 1;
+    // Blocks found in DB. Use the highest block num minus the block batch count.
+    // We need to reparse the previously parsed blocks because the blocks are added
+    // async and there may be blocks missing in the middle.
+    start = blocks[0].blockNum - BLOCK_BATCH_COUNT;
```
henryzbw: nit: might be btter to `start =Math.max(0, blocks[0].blockNum - BLOCK_BATCH_COUNT);`
may never meet, but better to handle it.

dwalintukan: good call. "should" be unreachable but better to handle it.

PR #286, opened 2019-07-15
Add address name service

henryzbw:
# Pull Request Template

## Checklist


src/api/address-name-service.js
```diff
@@ -0,0 +1,30 @@
+const { isUndefined, map, isArray } = require('lodash');
+const axios = require('axios');
+const { CONFIG } = require('../config');
+const { BLOCKCHAIN_ENV } = require('../constants');
+const logger = require('../utils/logger');
+
+const getNakaBaseUrl = () => {
+  const address = CONFIG.NETWORK === BLOCKCHAIN_ENV.MAINNET
```
dwalintukan: fyi, there is a `isMainnet()` func in config

src/config/index.js
```diff
@@ -19,6 +21,7 @@ const CONFIG = {
   DEFAULT_LOG_LEVEL: 'debug',
   STARTING_CONTRACT_VERSION_MAINNET: 5,
   STARTING_CONTRACT_VERSION_TESTNET: 0,
+  API_KEY: process.env.apiKey,
```
dwalintukan: all caps, process.env.API_KEY

src/graphql/queries/biggest-winners.js
```diff
@@ -78,6 +80,11 @@ module.exports = async (
     winnings = winnings.slice(skip, end);
   }
 
+  await eachOfSeries(winnings, async element => {
```
dwalintukan: add to schema and add as resolver

src/graphql/queries/leaderboard.js
```diff
@@ -31,6 +34,11 @@ const eventLeaderboardEntries = async (
     skip,
   });
 
+  await eachOfSeries(res.items, async element => {
```
dwalintukan: add to schema and add as resolver

src/graphql/queries/leaderboard.js
```diff
@@ -55,6 +63,11 @@ const globalLeaderboardEntries = async (
...
```
dwalintukan: add to schema and add as resolver

src/graphql/queries/most-bets.js
```diff
@@ -93,6 +94,11 @@ module.exports = async (
     items = items.slice(skip, end);
   }
 
+  await eachOfSeries(items, async element => {
```
dwalintukan: add to schema and add as resolver

src/route/api-address-name-service.js
```diff
@@ -0,0 +1,16 @@
+const { Router } = require('express');
+const { onRequestSuccess, onRequestError } = require('./utils');
+const AddressNameServiceApi = require('../api/address-name-service');
+
+const router = Router();
+
+router.get('/resolveAddress', (req, res, next) => {
```
dwalintukan: can delete this, we dont need to expose this API to the public. the UI shouldn't call this ever and all the names should be returned in all the graph queries.

src/route/api.js
```diff
@@ -1,9 +1,11 @@
 const logRoutes = require('./api-log');
 const configManagerRoutes = require('./api-config-manager');
 const multipleResultsEventRoutes = require('./api-multiple-results-event');
+const addressNameService = require('./api-address-name-service');
 
 module.exports = (app) => {
   app.use('/log', logRoutes);
   app.use('/config-manager', configManagerRoutes);
   app.use('/multiple-results-event', multipleResultsEventRoutes);
+  app.use('/address-name-service', addressNameService);
```
dwalintukan: remove

src/sync/result-set.js
```diff
@@ -28,7 +28,9 @@ const syncResultSet = async ({ startBlock, endBlock, syncPromises, limit }) => {
           // Parse and insert result set
           const resultSet = parseResultSet({ log: logObj });
           await DBHelper.insertResultSet(resultSet);
-
+          if (resultSet.eventRound === 0) {
```
dwalintukan: no need to check, this event is always round 0 result sets. `vote-result-set` is the ones that are voted over the threshold.

src/utils/index.js
```diff
@@ -26,4 +28,28 @@ module.exports = {
...
+   * Returns the lowercased string if the string is valid.
+   * @param {string} str String to lowercase
+   * @return {string} Lowercased string
+   */
+  getAndInsertNames: async (address, DBHelper) => {
+    if (isUndefined(address)) return;
+    let addresses = address;
+    if (!isArray(address)) addresses = [address];
+    const toCheckAddresses = [];
+    await eachOfSeries(addresses, async (item, index) => {
```
dwalintukan: this doesn't need to be in series since theres no limit to the number of queries for the db (at least that I know of). the ~40 req/second limit only applies to calls to the client server, ie. contract calls.

src/utils/index.js
```diff
@@ -26,4 +28,28 @@ module.exports = {
...
+      const ret = await DBHelper.findOneName({ address: item });
+      if (!ret) {
+        toCheckAddresses.push(item);
+      }
+    });
+
+    const addressNameObj = await resolveAddress(toCheckAddresses);
+    Object.keys(addressNameObj).forEach((key) => (addressNameObj[key] == '') && delete addressNameObj[key]);
```
dwalintukan: lodash filter might be more efficient?

src/utils/index.js
```diff
@@ -26,4 +28,28 @@ module.exports = {
...
+    await eachOfSeries(addressNameObj, async (name, key) => {
```
dwalintukan: same thing, DB operations can be in parallel with no limit.

src/utils/index.js
```diff
@@ -26,4 +28,28 @@ module.exports = {
     if (!isString(str)) return str;
     return str.toLowerCase();
   },
+
+  /**
+   * Returns the lowercased string if the string is valid.
```
dwalintukan: Check when copy/pasting

src/db/db-helper.js
```diff
@@ -579,4 +579,54 @@ module.exports = class DBHelper {
...
+      } else {
+        await DBHelper.updateName(nameEntry.address, nameEntry);
+      }
+    } catch (err) {
+      logger.error(`INSERT Name error: ${err.message}`);
+      throw err;
+    }
+  }
+
+  static async updateName(address, fields) {
```
dwalintukan: we dont need `update`. in the Naka ANS, you can only assign once and its permanent. so that means we have to be sure when we insert into the Nakabase DB that its good data.

test/db/index.js
```diff
@@ -2149,6 +2149,123 @@ describe('db', () => {
...
+        await DBHelper.insertName(newNameWithExistedAddress);
+        nameCount = await db.Names.count({});
+        nameCount.should.equal(3);
+        name = await DBHelper.findOneName({ address: newNameWithExistedAddress.address });
+        should.exist(name);
+        expect(name).excluding(['_id']).to.deep.equal(newNameWithExistedAddress);
+      });
+    });
+
+    describe('update name', () => {
```
dwalintukan: dont need update tests since we should never update an assigned name. this also means we should never insert an empty name.

test/db/index.js
```diff
@@ -2149,6 +2149,123 @@ describe('db', () => {
...
+        nameCount.should.equal(2);
+        await DBHelper.insertName(mockNames[2]);
+        nameCount = await db.Names.count({});
+        nameCount.should.equal(3);
+        name = await DBHelper.findOneName({ address: mockNames[2].address });
+        should.exist(name);
+        expect(name).excluding('_id').to.deep.equal(mockNames[2]);
+      });
+
+      it('should not insert existed address', async () => {
```
dwalintukan: remove, this test is obsolete. remove any update logic for `Names` since we will never do it.

src/db/db-helper.js
```diff
@@ -579,4 +579,54 @@ module.exports = class DBHelper {
...
+    }
+  }
+
+  static async insertName(nameEntry) {
+    try {
+      const existing = await DBHelper.findOneName({ address: nameEntry.address });
+      if (isNull(existing)) {
+        await db.Names.insert(nameEntry);
+      } else {
+        await DBHelper.updateName(nameEntry.address, nameEntry);
```
dwalintukan: remove update call

src/utils/index.js
henryzbw: don't know lodash filter how to filter like {"0xbc4b8726f9619c871fad66030116964480205b9d": "satoshi"}

dwalintukan: you should be able to use the iteration func, `filter(addressNameObj, (value, key) => addressNameObj[key]);` should filter out if the str is not empty

src/api/address-name-service.js
```diff
@@ -0,0 +1,29 @@
+const { isUndefined } = require('lodash');
+const axios = require('axios');
+const { CONFIG, isMainnet } = require('../config');
+const logger = require('../utils/logger');
+
+const getNakaBaseUrl = () => {
+  const address = isMainnet
```
dwalintukan: you have to call the func.

PR #293, opened 2019-07-19
Event tests fix

dwalintukan:
# Pull Request Template

## Checklist

## Notes

Fix tests after disabling them when pushing to master.


henryzbw (approved):
LGTM

PR #296, opened 2019-07-22
Api key env

dwalintukan:
# Pull Request Template

## Checklist


henryzbw (approved):
I added the .env to the path, please help check if its right.

dwalintukan:
@henryzbw yup its correct 👍

PR #299, opened 2019-07-22
Change to post

henryzbw:
# Pull Request Template

## Checklist

### DIFF
--- a/src/api/address-name-service.js
+++ b/src/api/address-name-service.js
@@ -14,9 +14,8 @@ module.exports = {
   async resolveAddress(address) {
     try {
       if (isUndefined(address)) throw TypeError('address is not defined');
-      const { data } = await axios.post(`${getNakaBaseUrl()}/resolve-address`, {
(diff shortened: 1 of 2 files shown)


src/api/address-name-service.js
```diff
@@ -14,11 +14,9 @@ module.exports = {
   async resolveAddress(address) {
     try {
       if (isUndefined(address)) throw TypeError('address is not defined');
-      const { data } = await axios.get(`${getNakaBaseUrl()}/resolve-address`, {
-        params: {
-          address,
-          apikey: CONFIG.API_KEY,
-        },
+      const { data } = await axios.post(`${getNakaBaseUrl()}/resolve-address`, {
```
BernardHan: nakabase api schema for post is different, you now need to include api key in the url.

`${getNakaBaseUrl()}/APIKEY/resolve-address`

Remove the apikey from the body

dwalintukan (approved):

PR #301, opened 2019-07-23
Fix migration 2 error

henryzbw:
# Pull Request Template

## Checklist


BernardHan:
Leaderboard entry investment wrong number can be fixed by this PR

dwalintukan (approved):

